PR #5, opened 2016-05-07
sortOn

matrix3d:


matrix3d:
float number

frameworks/projects/Core/src/main/flex/org/apache/flex/utils/Language.as
```diff
@@ -361,5 +370,56 @@ package org.apache.flex.utils
             return boundMethod;
         };
 
+		/**
+		 * @author lizhi
+		 * @param	arr
+		 * @param	names
+		 * @param	opt
+		 */
+		public static function sortOn(arr:Array,names:Object,opt:int=0):void{
```
adufilie: Sort options can also be an array if names is an array http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/Array.html#sortOn()

matrix3d: yes ,array work now.

matrix3d: https://github.com/matrix3d/spriteflexjs/blob/master/test/src/TestSortOn.as#L23

matrix3d: https://github.com/apache/flex-asjs/pull/6

matrix3d: the adobe flash player array.sorton ,opt array.
just use the opt[0],the other [1] [2],not work

PR #10, opened 2016-07-22
Binary data improvement

greg-dove:
As discussed with Harbs, this is my effort to get BinaryData closer to the regular flash ByteArray, particularly in terms of getting close to the IDataInput and IDataOutput flash interfaces, and fixing a few bugs that were only discovered through variations in use from testing. I have left out the amf write/read stuff and the multibyte text, but I think the rest is there now. There is at least one breaking change against the previous behavior : readByte() is now a signed byte, readUnsignedByte() is needed whereever that was being used.
I've added unit tests, which only run against the swf build, but I have manually tested and verified these all against the js output as well, across Chrome, Edge 13, IE11, Firefox.
Comments:
I think the data getter could be removed from BinaryData as on swf the array getter now provides the same thing and on js the underlying ArrayBuffer is also accessible on the Uint8Array from the array getter.
I added endianness detection to Endian with defaultEndian static getter. This should always resolve to one of the two types but I left an unresolved 'unknownEndian' state if something does not work here. This might be useful in terms of letting a server know what byte order the client prefers etc.
L

### DIFF
--- a/frameworks/projects/Core/src/main/flex/org/apache/flex/utils/BinaryData.as
+++ b/frameworks/projects/Core/src/main/flex/org/apache/flex/utils/BinaryData.as
@@ -18,59 +18,60 @@
 ////////////////////////////////////////////////////////////////////////////////
 package org.apache.flex.utils
 {
+
 COMPILE::SWF
 {
     import flash.utils.ByteArray;
 }
 
-    
+
 /**
  *  The BinaryData class is a class that represents binary data.  The way
  *  browsers handle binary data varies.  This class abstracts those
- *  differences..  
- *  
+ *  differences..
+ *
  *  @langversion 3.0
  *  @playerversion Flash 10.2
  *  @playerversion AIR 2.6
  *  @productversion FlexJS 0.0
  */
-public class BinaryData
+public class BinaryData implements IBinaryDataInput, IBinaryDataOutput
 {
     /**
      *  Constructor. The constructor takes an optional bytes argument.
      *  In Flash this should be a ByteArray. In JS this should be an ArrayBuffer
-     *  
+     *
      *  @langversion 3.0
      *  @playerversion Flash 10.2
      *  @playerversion AIR 2.6
      *  @productversion FlexJS 0.0
      */
     COMPILE::SWF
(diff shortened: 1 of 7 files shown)


Harbs:
I gave this a quick look-over. The only issue that I noticed was that the default behavior of Endianess is different between Flash and JS.

In Flash, the default is bin endian, while in JS, the default is machine dependent which over 90% of the time means little endian.

That's why I made the default little endian (which is different than the default Flash behavior). There's no really "right" answer to this question, so I'm not totally sure which approach is best.

adufilie:
The default should be big endian ("network byte order") used by java and flash.

Harbs:
The thing is: Endianness based on the hardware was done for a reason:
http://calculist.org/blog/2012/04/24/the-little-endian-web/

Basically, if the data is being used for GPU drawing, it needs to be native for efficiency. I'm not sure if BinaryData objects will be used for that purpose, but it's something to keep in mind.

aharui:
Someday, we can have a BinaryDataForGPU class with native defaults.  Or make more classes like BinaryDataLittleEndian and BinaryDataBigEndian.  No need to figure it out all in a single class.

adufilie:
Thanks for the link, Harbs. I mistakenly thought this code was only related to ByteArray.  Now I see that it's actually writing to an ArrayBuffer in JavaScript.

greg-dove:
I just read back on the other comments, some of which I had missed seeing
before my earlier reply. If defaulting to big endian is necessary on js for
consistency, that's easy simply by changing the initialize value in
BinaryData and everything else still works as described above (explicitly
setting to LITTLE_ENDIAN later will still use optimized code on js).

greg-dove:
Harbs, I have a number of improvements, including performance improvements, and one bug fix (writeUTF was not respecting the endian setting for the prepended string length) that supersede this request. Please feel free to close this pull request, I will issue another tomorrow with the improvements..

Harbs:
Sounds good. I was going to merge this today, but I'll wait for your updates.

The only way for me to close pull requests is to add a commit (because this is just a mirror of the Apache repo). I think you should be able to close it though.

PR #11, opened 2016-07-26
Binary data improvement

greg-dove:
As per recent discussions - updated BinaryData improvements
-this request fixes a bug in writeUTF that was not respecting the endian setting for the prefixed short value.
- BinaryData now defaults to BIG_ENDIAN for consistency with flash:
substantial performance improvements were achieved on Chrome (Windows 10) which mean that big endian is now as fast as little endian, which itself is also a little faster than before. (a test was done with 32bit unsigned ints- 2 separate loops of 100,000 with write and read in each, using preallocated buffers).  The alternate endian approach swaps out DataView with bitwise operators to rotate the byte order when necessary on read and write. In terms of other browsers, there was no obvious change on IE 11 and a slight degradation on Edge, FireFox had a degradation also. But the selection was made for bias towards performance on Chrome based on desktop browser market share. Big endian ops on Chrome are approximately 2.85x faster using the bitwise operator approach and virtually indistinguishable from little endian mode.
Endian.systemEndian provides the detection approach for detecting system/hardware endian (on js).
Tests and doc comments were updated also.


Harbs:
Thanks for this!
I just merged your changes.

PR #18, opened 2017-10-03
Update Notice to include relevant parts from FlexSDK NOTICE file

justinmclean:
As FlexSDK is ALv2 relevant bits in it's notice file need to be propagated to this notice file. See [1]
1. http://www.apache.org/dev/licensing-howto.html#alv2-dep

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -3,3 +3,10 @@ Copyright 2012-2017 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+The Initial Developer of the Original Code, known as Adobe Flex, is Adobe
+Systems Incorporated (http://www.adobe.com/).
+    Copyright 2003 - 2012 Adobe Systems Incorporated. All Rights Reserved.


Harbs:
Which dependency do you think requires this change?

I can't think of a Flex dependency in asjs that would require this.

justinmclean:
It's not a dependancy issue. You copied code from the Flex SDK into the Royale SDK. The Flex SDK is ALv2 licensed and it has a NOTICE file so bits that are important in that NOTICE file need to be copied the the Royale NOTICE file to comply with the Apache license and ASF policy.

Harbs:
I assumed that if you linked to requirements for dependencies you were talking about a dependency requirement.

The FlexSDK had those bits because it was a product that came from Adobe. Considering that royale-asjs is a project developed solely within Apache, I don't see a need to include that here, and the text is actually not factually correct.

justinmclean:
I assume you would agree that the MX Validation class (which have been copied into this branch from the Flex SDK) were written by people at Adobe?

dave2wave:
Two points.

(1) The PR has an extra copyright for the ASF added. This would need to be removed from the PR.

(2) The inclusion of the following:

The Initial Developer of the Original Code, known as Adobe Flex, is Adobe Systems Incorporated (http://www.adobe.com/).
    Copyright 2003 - 2012 Adobe Systems Incorporated. All Rights Reserved.

This is debatable, It is debatable in both Flex and Royale. I think that the debate is more about the inclusive “The Initial Developer …”

Better language would be something like:

“Includes software developed and distributed as Adobe Flex.
    Copyright 2003 - 2012 Adobe Systems Incorporated. All Rights Reserved.”

Harbs:
I agree on both points.

I'd personally be okay with the "better language" although I don't see the need to include anything.

aharui:
The NOTICE files in other Royale repos already contains similar "original developer" language and try to distinguish that not all of the code base was originally developed by Adobe.

IMO, we should try to quickly agree on language to make that same distinction for the royale-asjs repo.  I suggest:

> The Initial Developer of the some parts of the framework, which are copied, derived or inspired by Adobe Flex (via Apache Flex), is Adobe Systems Incorporated (http://www.adobe.com/).
Copyright 2003 - 2012 Adobe Systems Incorporated. All Rights Reserved.

I suggest this because I would like to avoid spending energy listing out every single place that was somehow influenced by Adobe code, give Adobe proper attribution per the Flex NOTICE language, and distinguish that lots of code was written at Apache.

Dave's "includes" suggestion sounds more like a bundling which doesn't feel right to me.

dave2wave:
Alex your language should be fine.

Harbs:
There's an extra "the" in the first line:
>The Initial Developer of **_the_** some parts of the framework

Otherwise the wording looks fine to me.

aharui:
Removed the extra "the" and added two "from" and a comma.

Issue #23, opened 2017-10-09
Can select empty cells in DateChooser


justinmclean:
Clicking on empty cells in DateChooser selects the cell and set the selected date to null this can cause RTEs.

Sample code:
```
<?xml version="1.0" encoding="utf-8"?>
<js:Application xmlns:fx="http://ns.adobe.com/mxml/2009"
                xmlns:js="library://ns.apache.org/flexjs/basic">

    <js:valuesImpl>
        <js:SimpleCSSValuesImpl/>
    </js:valuesImpl>

    <fx:Script>
        <![CDATA[
        public function displayDate():void {
            display.text = chooser.selectedDate.toDateString();
        }
        ]]>
    </fx:Script>

    <js:initialView>
        <js:View>
            <js:Group id="content">
                <js:beads>
                    <js:VerticalLayout />
                </js:beads>
                <js:DateChooser id="chooser" width="300" click="displayDate()" />
                <js:Label id="display"/>
            </js:Group>
        </js:View>
    </js:initialView>

</js:Application>
```

To reproduce just click on an cell without a number in it.

PR #25, opened 2017-10-09
Only dispatch change event if new date is selected. Fixes #24 and partial fix for #23.

### LINKED ISSUE #24: Change event dispatched by DateChooser when same date is selected
Clicking on the same date twice will cause two change events to be dispatched by DateChooserMouseController.

justinmclean:
Stops potential RTE in client code when clicking on empty date cell and stops change event being dispatched if the same date is clicked on twice.


Harbs:
The `change` event should never have been dispatched from here.
Fixed in 2072541

PR #38, opened 2017-10-18
When changing minimum or maximum checks that value is not out of range. Fixes #37

### LINKED ISSUE #37: NumericStepper will show zero even when initial value is above zero
If you set the minimum value (like 1) to a value the model gets updated but the existing value doesn't and stay at the default zero.

Code to reproduce:
```
                xmlns:js="library://ns.apache.org/flexjs/basic">


    <js:valuesImpl>
        <js:SimpleCSSValuesImpl/>
    </js:valuesImpl>

    <js:initialView>
        <js:View>
            <js:Group id="content">
                <js:beads>
                    <js:VerticalLayout />
                </js:beads>
                <js:NumericStepper id="ns" minimum="1" maximum="10" stepSize="1" />
            </js:Group>
        </js:View>
    </js:initialView>

</js:Application>
```
Even though the minimum is set to 1 0 is displayed like so:

justinmclean:


Harbs:
This does not belong in the RangeModel. Models (especially globally used ones like RangeModel) should not have just in case code like this.

If the value should be set to be within the range, that should be done by code which uses the model.

PR #54, opened 2017-10-26
Date chooser performance

justinmclean:
Improved performance by 20% or so as there's no need to work out the _days array length 42 times as the length doesn't change. 

Tested on Chrome so performance improvement may be different on other platforms.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1718,4 +1718,10 @@
             <arg value="${file}" />
         </java>
     </target>
+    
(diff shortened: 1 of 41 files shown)


Harbs:
It turns out that there was no need for three loops at all.

It looks like someone had started optimizing this class by initializing the array with a length of 42. That makes the first and last loop unnecessary.

I simplified the middle loop into a simple while loop. There was no need for two conditions.

PR #55, opened 2017-10-26
Date chooser no year check

justinmclean:
As the calendar doesn't span months their is no need to check if the years are the same or oddly that the year is non zero as the code was currently doing.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1718,4 +1718,10 @@
             <arg value="${file}" />
         </java>
     </target>
+    
+    <target name="print.geckodriver.version" >
+        <exec executable="${webdriver.gecko.driver}" failonerror="true">
+            <arg value="--version"/>
+        </exec>
+    </target>
 </project>
--- a/examples/build_example.xml
+++ b/examples/build_example.xml
@@ -167,7 +167,7 @@
         <!-- Load the <compc> task. We can't do this at the <project> level -->
         <!-- because targets that run before flexTasks.jar gets built would fail. -->
         <taskdef resource="flexTasks.tasks" classpathref="lib.path"/>
-        <property name="config_arg" value="flex" />
+        <property name="config_arg" value="royale" />
         <property name="file_suffix" value="mxml" />
         <property name="optional_arg" value="-define+=CONFIG::dummy,true" />
         <mxmlc fork="true"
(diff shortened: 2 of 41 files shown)


Harbs:
The year comparison is necessary because the month of the following year can be displayed.

The code seemed to be missing the second half of the year equality check.

toDateString() and caching the _selectedDate date string is a probably a more efficient way of doing this check and definitely simpler.

justinmclean:
One one month at a time is displayed so how can the note of the following year be displayed? If the month of the following year is displayed then the month has changed to 1 from 12 so the year check is not required. Please merge the pull request.

Harbs:
This is a moot point because the change I made makes the whole question an academic one, but, to answer your point:

The selectedDate should not change when displaying a different month. Imagine October 29 2017 is selected. The user then pages through the months to October 2018. This does not change the selected date. The logic needs to differentiate between October 29 2017 and October 29 2018 and not display October 29 2018 as the selected date.

There appears to be a bug in the display where the selected date is not shown when coming back to that month, but that is a separate issue.

justinmclean:
If want to include the year check then this would be a lot more efficient than the original code.
```
        public function getIndexForSelectedDate():Number
        {
            var d:Date = _selectedDate;
            if (d == null || _days[0].getFullYear() != d.getFullYear() || _days[0].getMonth() == d.getMonth()) return -1;
            
            for(var i:int=0; i < _days.length; i++) {
                var test:Date = _days[i] as Date;
                if (test != null) {
                    if (test.getDate() == d.getDate())
                        return i;
			    }
            }
            return -1;
		}
```

As only a single month is display at a time then you only need to check the year and month once.

Harbs:
That’s not going to work because I believe _days[0] can be null.

There might be a way to make the check a little more performant, but I can’t think of a way to do so without making the code convoluted.

If you are worried about the performance of getDateString() vs separate calls to getMonth(), getDate(), etc, you can create a performance test on https://jsperf.com and we can see what we’re dealing with.

I have other higher priority issues at the moment, but we don’t want to introduce bugs trying to make theoretical performance improvements. My quick tests seem to indicate that toDateString() is pretty quick. I'm not sure my tests are accurate, but if they are, it looks like 30 checks for toDateString() are considerably less than a millisecond.

Issue #50, opened 2017-10-25
Update to select a range of dates in DateChooser


justinmclean:
Wen trying to select a range of dates the text gets selected but only the last date is selected this is confusing.
![2017-10-25_18-46-58](https://user-images.githubusercontent.com/144504/31986716-370078ac-b9b5-11e7-847e-0f07546d4e88.png)

This is a common use case e.g. selecting the check in and check out dates for a hotel.

Visually it may be best to disable text selection in the date chooser?

The ability of selecting a date range  would need to be added via a new bead rather than the existing component.

pentapache:
Probably best to prevent selection and should be an easy fix. Range selection might be accomplished by adding a bead.

PR #58, opened 2017-10-28
Make dates in DateChooser unselectable

justinmclean:
Stops ugly visual effect when multiple dates are selected in a DateChooser partial fix for #50 as suggested by @pentapache 
Also see http://caniuse.com/#feat=user-select-none about 95% of browsers support this so probably no need for other non standard browser prefixes

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1718,4 +1718,10 @@
             <arg value="${file}" />
         </java>
     </target>
+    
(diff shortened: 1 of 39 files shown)


Harbs:
This was fixed in 4ccfc82 which has more browser support.

PR #63, opened 2017-11-01
Fix #62 selected date not being displayed

### LINKED ISSUE #62: DateChooser displayed selected date lost when changing month
To reproduce:
1. select a date
2. click on next month
3. click on previous month

The selected date doesn't show when it should.

justinmclean:


Harbs:
There is no code in the pull request and as far as I can tell, no bug to fix.

PR #65, opened 2017-11-03
DateChooser set first day of week

justinmclean:


justinmclean:
Updated with new pull request #98

PR #75, opened 2017-11-13
Set all padding values at once

justinmclean:


justinmclean:
Example of use:
```
    
    <js:valuesImpl>
        <js:SimpleCSSValuesImpl/>
    </js:valuesImpl>

    <js:initialView>
        <js:View>
            <js:style>
                <js:SimpleCSSStyles padding="20" />
            </js:style>
            <js:Container width="100%">
                <js:beads>
                    <js:HorizontalLayoutWithPaddingAndGap padding="10" gap="20" />
                </js:beads>
                <js:TextButton text="Button 1" />
                <js:TextButton text="Button 2" />
                <js:TextButton text="Button 3" />
             </js:Container>
        </js:View>
    </js:initialView>

</js:Application>
```

bigosmallm:
Justin, your PRs have too many commits.  Please squash them into 1 commit per PR.  That will make it easier for folks to review your changes.

justinmclean:
Click on the file changes and you can easily see what's changed. AFAICS squash commits wont help here as the other commits have been made in my develop branch not in this branch.

Harbs:
@bigosmallm FWIW, you can view the file changes. That makes it easier.

I'm not sure if this change is PAYG. It's a nice convenience to be able to specify all padding at once, but not strictly necessary and adds bulk to the bead.

It feels to me like this belongs in an Express component or similar. An express-type component can allow specifying padding to the component itself which is even easier than having to specify the layout.

justinmclean:
If the function is not called it will be optimised out so there is no PAYG cost here if you don't use it. Not having it is likely to increase code size as you need to specify all four attributes and values values rather than just one if you want the same (which is fairly common IMO).

Also SimpleCSSStyles has a padding attribute and that hasn't been called out as not PAYG.

bigosmallm:
Commit history is part of the PR.  If we merge it as is, it will create unnecessary git history.  The committer who is merging the PRs should remember to `squash and merge`.  
Justin perhaps you should create a new branch for each PR?

Harbs:
SimpleCSSStyles was created to support the styles that are supported by Flash. Different story there.

I'm currently on the fence on this issue. I'd like to hear what others think.

piotrzarzycki21:
Express or separate bead for setting all paddings at once . Agree with you @Harbs .

justinmclean:
@bigosmallm  Each PR is on a new branch off develop the issue is outstanding PR that have not been accepted.

justinmclean:
Also squashing commit is considered (IMO) bad practice at Apache as it changes history. I don't think we've ever had a discussion about this so please bring it up on the dev list. That being said it's usually not an issue as after a few PRs the person is usually made a committer.

bigosmallm:
23 commits for a 8 line change is excessive.  I'm closing this one as Invalid.

justinmclean:
There is only one commit in the branch in question please reopen and merge

justinmclean:
You'll see that most of the merges above are "Merge remote-tracking branch 'upstream/develop' into develop" which I need to do in order to keep my develop branch in sync with the upstream apache repo.

justinmclean:
The change was a single comment on a branch off my develop as you can see here. [1]

1. https://github.com/apache/royale-asjs/pull/75/commits/c2ab730e2d2e73703b4361aea4bdf20b6c1cc02d

justinmclean:
So I tried to fix this and was unable to (hard reset develop to upstream, re-parent branch and rebase). However you should be able to apply the single commit like so:
git fetch --all
git cherry-pick c2ab730

You may need to git remote add my repo I'm not sure it's been added and as I don't have access I can't test it.

Alternatively you could just take the patch file and just apply the last commit in it. [1]

1. https://patch-diff.githubusercontent.com/raw/apache/royale-asjs/pull/75.patch

Or I can submit another pull request if you prefer.

justinmclean:
Also re PAYG:

With left , right, top and bottom padding version the example program release size is 83,026 bytes
With single padding version the example program release size is 82,960 bytes.

So using the single padding attribute will save you a few bytes - 66 in this case. Not significant but it doesn't add to the size so there's no PAYG cost there.

Again if the padding attribute is not used then the function is optimised out / removed so there's no PAYG size cost if you don't use the single padding attribute.

Just out of interest I looked at a mixed version:
1 single padding and 1 set of padding attributes 83,082 bytes.
2 sets of padding attributes 83,148 bytes.

So even if you mix them there is a few bytes saved and no PAYG cost.

Runtime wise the left right, top and bottom version it needs to make four function calls to set the padding. The single padding version will be faster as it only calls one function. Not that you could probably measure it in any meaningful way or that it would have a noticeable impact on a simple application but it may provide a small benefit in complex layout, item renderers or the like.

There is no runtime PAYG cost if you use or don't use the single padding attribute.

So can you please explain how you find this not PAYG?

dave2wave:
My two cents here is that having patterns that fit into the CSS / HTML way make sense. It should also make it easier to convert HTML/JS developers over to Royale.

justinmclean:
In CSS you can go "padding: 25px;" and that will apply padding to top, bottom, left and right. The added padding attribute matches that.

bigosmallm:
Before adding a method, ask yourself if the user can opt-out of this functionality if they don't need it.  
The setters for left,right,top and bottom are required functionalities.  The set_all_padding method is a convenience function that not everyone is going to use.  

The correct comparison would be to look at the file size with and without the set_all_padding method.  

Convenience methods should be added into higher order components (for example Express) or in custom beads.  That way, the functionality is optional.

justinmclean:
I did there's is no PAYG size cost their either it's 83,066 bytes.

I seem to repeating myself here remember that unused methods are optimised away / removed by the compiler in the release version.

PR #78, opened 2017-11-13
Stop text overflow in Button fixes #77

### LINKED ISSUE #77: Button text can extends outside of button
This example show that button text can extend outside of the button.
```
    
    <js:valuesImpl>
        <js:SimpleCSSValuesImpl/>
    </js:valuesImpl>

    <js:initialView>
        <js:View width="100%" height="100%">
            <js:Container width="100%" height="100%">
                <js:beads>
                    <js:HorizontalLayoutWithPaddingAndGap gap="20" />
                </js:beads>
                <js:TextButton text="A Button with a lot of text on it" width="100" />
                <js:TextButton text="A Button" width="100" />
             </js:Container>
        </js:View>
    </js:initialView>

</js:Application>
```

Would be nice to have a truncate text bead (or similar) to stop this from happening.

justinmclean:


dave2wave:
On the mailing list it was requested that this PR be reduced to a single commit

Harbs:
Reducing it will not fix this PR.

Besides the extra function which should have been cleaned up, the types are wrong in JS blocks.

justinmclean:
I didn't see that request on the mailing list. In order to make it a single commit I'll need to make a new PR that OK?

justinmclean:
@Harbs can you be a little more precise. Which function are you referring too and which types are "wrong"? This has been tested and is being used in an application btw.

Harbs:
1. An `updateHost` function which takes null is simply sloppy. I'm not sure why it's an extra function at all. The code could easily have been inlined. Even so, a function should not have a required parameter which is only ever called with null.
2. IUIBase does not have an `element` getter.

Harbs:
Correction: I was wrong on point 2. `IUIBase` does inherit an `element` getter from `IRendererObject` (through `IChild`) Sorry about that.

justinmclean:
@Harbs  re your first point I was just following the pattern as done in other parts of the SDK for instance see 1 which also does calls updateHost passing null.

1. https://github.com/apache/royale-asjs/blob/d407bcf8e5a0a1ad23108716ccac63ab28c58f08/frameworks/projects/Basic/src/main/royale/org/apache/royale/html/beads/DisabledAlphaBead.as

justinmclean:
simplified pull request see #97

PR #85, opened 2017-11-14
Clicking on track set value to NaN. Fixes #83

### LINKED ISSUE #83: Slider value is set to NaN if track is clicked on
Clicking on the track sets the value of the slider to NaN.

Also track get focus / selection indicator should it?

Example code:
```
   <js:valuesImpl>
       <js:SimpleCSSValuesImpl/>
   </js:valuesImpl>

    <js:beads>
        <js:ApplicationDataBinding />
    </js:beads>

   <js:initialView>
       <js:View width="200" height="200">
           <js:Container>
               <js:beads>
                   <js:VerticalLayout />
               </js:beads>
               <js:Slider id="s" minimum="0" maximum="10" height="50" />
               <js:Label text="{s.value}" />
           </js:Container>
       </js:View>
   </js:initialView>
</js:Application>
```

justinmclean:
Code assumed thumb was being dragged and not that the track was clicked on and in that case offsetX is undefined.


justinmclean:
Slider code has been rewritten and this PR no longer applies.

PR #88, opened 2017-11-14
Adjust alert title bar size so that it doesn't go over it's RHS border

justinmclean:
The 98% is correct as the width is hardcoded to 200 pixels wide. If the width was to change then the percentages also also need to change.


Harbs:
This is not the right way to resolve the problem. The issue was caused by box-sizing defaulting to content-box since the Alert was outside the view.

Fixed in 21cbd13

Issue #93, opened 2017-11-28
No easy cross platform way to set or unset focus on a Button


justinmclean:
There is no API or bead to set or unset focus on a Button.

You can do this in JS:
```
            COMPILE::JS {
                button.element.blur();
            }
            COMPILE::JS {
                button.element.focus();
            }
```

justinmclean:
Pull request #93 contains a focus bead and adds it to the express TextButton.

Sample use:
```
<?xml version="1.0" encoding="utf-8"?>
<js:Application xmlns:fx="http://ns.adobe.com/mxml/2009"
                xmlns:js="library://ns.apache.org/royale/express"
                applicationComplete="init()">

    <fx:Script><![CDATA[
        public function init():void {
            important.focus();
        }
        ]]></fx:Script>

    <js:valuesImpl>
        <js:SimpleCSSValuesImpl/>
    </js:valuesImpl>
    
    <js:initialView>
        <js:View percentWidth="100" percentHeight="100">
            <js:HContainer>
                <js:TextButton text="One"  />
                <js:TextButton text="Two"  />
                <js:TextButton id="important" text="Default" />
            </js:HContainer>
        </js:View>
    </js:initialView>

</js:Application>

```

greg-dove:
reopened, because I closed the wrong issue

PR #94, opened 2017-11-28
Add focus bead for TextButton

justinmclean:
Add FocusBead and focus and blur methods to express TextButton using the bead


justinmclean:
Fixes issue #93

PR #96, opened 2017-12-13
Button text overflow bead

justinmclean:
Bead to add "..." to text if it overflows


justinmclean:
Something went wrong there closing and trying again.

PR #98, opened 2017-12-13
Fix setting first day of the week fixes #21

### LINKED ISSUE #21: DateChooser firstDayOfWeek is non functional
Most countries don't use Sunday as the first day of the week. Only USA, Canada, Japan and a few other do so. Setting the day of the week to be Monday like so doesn't see to have any effect.

`model.firstDayOfWeek = 1;`

Looking at the setter it seem to me that it should be calling updateCalendar() (which it isn't) and perhaps changing model.dayNames to reflect the new order.

Even when dayName is modified and updateCalendar is called manually there is still no effect. It look like updateCalendar is missing code to take into account the first day of the week.

justinmclean:


Harbs:
This appears to break the functionality of setting `dayNames`.

carlosrovira:
Hi, didn't see this. Just Let you know that Jewel DateField and DateChooser has all this currently working. I think if use Basic, could borrow some fixes from Jewel

PR #53, opened 2017-10-26
Have date chooser dispatch month change events

justinmclean:
Example code:
<?xml version="1.0" encoding="utf-8"?>
<js:Application xmlns:fx="http://ns.adobe.com/mxml/2009"
                xmlns:js="library://ns.apache.org/royale/basic">

    <fx:Script><![CDATA[
        public function monthChange():void {
            trace("month changed");
        }
        ]]></fx:Script>

    <js:beads>
        <js:ApplicationDataBinding />
    </js:beads>

    <js:valuesImpl>
        <js:SimpleCSSValuesImpl/>
    </js:valuesImpl>

    <js:initialView>
        <js:View>
            <js:Group id="content" height="200" width="250">
                <js:beads>
                    <js:VerticalLayout />
                </js:beads>
                <js:DateChooser id="dc1" height="200" monthChanged="monthChange()" />
            </js:Group>
        </js:View>
    </js:initialView>

</js:Application>

PR #100, opened 2017-12-13
Data chooser month change event

justinmclean:


justinmclean:
Simpler version of PR #53. Didn't change the other event names.

Issue #102, opened 2017-12-13
Odd click behaviour with slider


justinmclean:
To reproduce use this code:
```
<?xml version="1.0" encoding="utf-8"?>
<js:Application xmlns:fx="http://ns.adobe.com/mxml/2009"
               xmlns:js="library://ns.apache.org/royale/basic">
   <js:valuesImpl>
       <js:SimpleCSSValuesImpl/>
   </js:valuesImpl>

    <js:beads>
        <js:ApplicationDataBinding />
    </js:beads>

   <js:initialView>
       <js:View width="200" height="200">
           <js:Container>
               <js:beads>
                   <js:VerticalLayout />
               </js:beads>
               <js:Slider id="s" minimum="0" maximum="1000" snapInterval="5" width="500" />
               <js:Label text="{s.value}" />
           </js:Container>
       </js:View>
   </js:initialView>
</js:Application>
```

Click on the middle of the track and the thumb will move there and have a value around 500, click again and the thumb moves right, click again and it will move right again. I would of expected to stay in the same place and not move right on any of those clicks.

pentapache:
This should be fixed by commit 02f0dd4

pentapache:
This should be fixed by commit 02f0dd4

PR #104, opened 2017-12-13
Using thumb width in calculations causes thumb to jump right

justinmclean:
When clicking on the thumb half the thumb width is added to the thumb for a simple click on a thumb (with no drag) the thumb should stay where it is.


justinmclean:
fixes #102

justinmclean:
not tested but I assume similar changes would also need to be made to calcValFromMousePosition in VSliderMouseController

justinmclean:
modify PR as calcValFromMousePosition in VSliderMouseController had some issue

PR #141, opened 2018-04-02
Copied MX Panel from flex-sdk

alinakazi:
Copied MX Panel from flex-sdk to Start mx.containers.Panel Emulation


aharui:
I did Panel

PR #148, opened 2018-04-06
Sort and SortField Added

alinakazi:
Commented :
import mx.resources.IResourceManager;
import mx.resources.ResourceManager;


aharui:
Alina,

Please comment out or delete any APIs and code that are not being used by your application.  We have enough other code paths to migrate that we don't want to take on things we don't need to.

aharui:
According to your API Report, some APIs in Sort and SortField like "unique", "reverse" and others are not being used by your app.  Those APIs should be commented out for now so there is less code that we have to get working.

Sort and SortField are being used by your application so it is fine to have these emulated, but we should not bring over APIs in those classes that are not being used.

PR #153, opened 2018-04-09
EffectEvent and IEffectInstance Added

alinakazi:
mx.effects.IEffectInstance and mx.events,EffectEvent  added


aharui:
Is there a reason PropertyChanges.as is in this set of files?  I don't see a need for it.

alinakazi:
PropertyChanges.as File Deleted & Reference removed from IEffectInstance.as and MXRoyaleClasses.as

PR #155, opened 2018-04-09
SandBoxMouseEvent and ResizeEvent

alinakazi:
SandBoxMouseEvent and ResizeEvent


aharui:
Again, there were some fields that aren't being used, but I will accept this for now.

PR #158, opened 2018-04-10
StringUtil.as

alinakazi:
mx.utils.StringUtil.as


aharui:
@piotrzarzycki21 We will take duplication to eliminate number of changes to existing Flex source.

PR #159, opened 2018-04-11
mx.graphics.GradientEntry Added

alinakazi:
Graphics->GradientEntry Added


Harbs:
Oops. The parameters are not being passed into the super constructor. I did not notice when I merged. Can you fix that?

PR #168, opened 2018-04-12
LineChart Added

alinakazi:
LineChart Added


aharui:
LineChart appears to be missing stubs for the APIs used according to the api-report.  It also appears from the api-report that LineChart and other charts will have to extend ChartBase.

PR #169, opened 2018-04-12
classfactory Added

alinakazi:


aharui:
ClassFactory should have an API called "properties".  See the api-report.

aharui:
Properties should be public, not private.  I would just re-copy the property from the original flex-sdk ClassFactory.as

PR #170, opened 2018-04-12
BarChart ColumnChart and PieChart Added

alinakazi:


aharui:
These charts, like LineChart, may need to subclass ChartBase and have APIs as listed in the api report.

PR #173, opened 2018-04-13
Move.as Added

alinakazi:


aharui:
This may need to be refactored on top of mx.effects.Tweeninstance some day, but we'll accept this for now.

PR #176, opened 2018-04-13
Sequence.as

alinakazi:
Dependent on IEffectInstance


aharui:
Isn't this missing IEffectInstance?

alinakazi:
Yes I mentioned that dependency in the comment above.

PR #217, opened 2018-04-26
ISort,ArrayCollection and ListCollectionView Added

LHR07-DBz:


aharui:
mx.collections.ArrayList will be emulated at some point and that code can be re-enabled.  It is ok to comment out things you think might be needed later.  The main goal is to emulate the classes in the API report so an app can compile, then worry about making them actually work.

PR #221, opened 2018-05-04
Added few chart classes

LHR07-DBz:
mx.charts.chartClasses.CartesianChart
mx.charts.chartClasses.IAxis
mx.charts.PlotChart
mx.charts.series.PieSeries
mx.charts.series.PlotSeries
mx.charts.series.items.PieSeriesItem
mx.charts.AreaChart


LHR07-DBz:
CartesianChart removed but IAxis is required as argument for horizontalAxis method so it is retained. Rest all files are updated as per comments

alinakazi:
Like setStyle & width are properties/methods required in PlotChart  and these are extended from UIComponent. IMO PlotChart should extend UIComponent. Again this is just an example. 

@aharui please comment on it further.

aharui:
Alina is correct that just about every class will need to extend something.  However, in this case PlotChart should extend ChartBase.  The way to determine this is:
1) Look at the flex-sdk version of the class (in this case PlotChart.as)
2) See if the base class (in this case CartesianChart) is listed in the API report.  If so, use that base class, if not, open the flex-sdk version of the base class (in this case CartesianChart.as)
3) See if its base class (in this case ChartBase) is listed in the API report).  In this case, it is, so PlotChart should extend ChartBase.

Loop through 2 & 3 until you find a base class in the report.  The loop for ui controls should end with UIComponent if not some subclass of UIComponent.

LHR07-DBz:
Updated AreaChart.as, PlotChart.as, PlotSeries.as, PieSeries.as & PieSeriesItem.as files as instructed

alinakazi:
Do not copy ChartBase from flex-sdk & make changes from beginning .  
Add/Copy methods in existing ChartBase that is already emulated.

LHR07-DBz:
Moved references from MXRoyaleClasses.as to mx-manifest.xml

PR #228, opened 2018-05-14
Added Validators

LHR07-DBz:
mx.validators.NumberValidator
mx.validators.PhoneNumberValidator
mx.validators.DateValidator
mx.validators.EmailValidator


alinakazi:
As you update these classes now these classes extends Validator. So the properties/methods that you copied from Validator delete them.  e.g. _property and its getter/setter are extended from Class Validator. So delete them from NumberValidator & others .

LHR07-DBz:
Modified as per comments

LHR07-DBz:
Moved references from MXRoyaleClasses.as to mx-manifest.xml

PR #239, opened 2018-05-18
Update AMFBinaryData.as

doug777:
Add new variable "flexTraitFound" and change readScriptObject() to allow ArrayCollection and ObjectProxy objects sent from CF to be correctly decoded. This change should not affect anything other than this. (Note that ObjectProxy is not tested, so a further change could be necessary and I will do this when I can test it with an ObjectProxy.)


carlosrovira:
Thanks Doug, your first PR is merged! Congratulations! Hope you bring more valuable contributions like this to Apache Royale! :)

PR #242, opened 2018-05-18
Base64Encoder, Base64JSWrapper, TextEncoderLiteWrapper classes added

nihavend:
Base64Encoder modified for emulation project and  Base64JSWrapper, TextEncoderLiteWrapper javascript classes added


nihavend:
@piotrzarzycki21 may you pls check the PR, some of your comments are not included because i need more advices.

piotrzarzycki21:
The only comment which I have is that. Shouldn't we have those encoder in Core or Basic ? Is it specified only to Emulation? @aharui

aharui:
Sure, at some point we should refactor to have a Basic version.  It isn't obvious to me how to do it, so I think we should accept the PR as-is and find a volunteer to refactor later.

PR #254, opened 2018-05-27
Base64Decoder class added and wrapper classes modified to support decoding

nihavend:


nihavend:
@piotrzarzycki21 and @aharui  hope you review PR.

nihavend:
Thanx @piotrzarzycki21, waiting for the merge :)

aharui:
@piotrzarzycki21 If you think it is good, feel free to merge it.  No need to wait for me.

frameworks/projects/MXRoyale/src/main/royale/mx/utils/Base64Decoder.as
```diff
@@ -0,0 +1,258 @@
...
+        result.position = 0;
+        
+        filled = 0;
+        return result;
+    }
+
+    COMPILE::SWF
+    public function flush():ByteArray
+    {
+        if (count > 0)
```
piotrzarzycki21: If we are not using that condition - Remove it completely

nihavend: @piotrzarzycki21  After commenting out IResourceManager and ResourceManager - i am really not sur to handle - the condition became meaningless. Sorry, i missed that. Will fix it.

piotrzarzycki21: I just noticed that I do not understand above comment fully. Are you saying that there can be scenario that you un-comment that code ? If NOT let's remove it instead commenting it and leave it in source code.

nihavend: I tought that i have to make the class can be compiled without error for swf and functional for JS. From your comments, i see that it is not a correct approach. I can remove flush method from the class instead of commenting but after, i should remove the functions calling flush and so on. Do you want me to go that direction ? Just to make it clear and do not let missunderstanding.

piotrzarzycki21:
Serkan maybe let's leave it as is. I willerge it and we can always fix that small issues later on.

PR #281, opened 2018-09-01
jewel validators

yestaro:
Hi, here are action script  classes of jewel validator, 

they need style as below

```css
.jewel.textinput.errorBorder input {
    border: 1px #EC1C24 solid;
}

.jewel.errorTip {
    padding: 6px;
    box-shadow: 0px 3px 3px 0 rgba(0, 0, 0, 0.4);
    color: #fff;
    background: #EC1C24;
    box-sizing: border-box;
    animation: slide-up 0.4s ease;
    z-index: 1;
    margin-top: -31px !important;
}

.jewel.errorTip::before {
    border: solid 8px transparent;
    top: 100%;
    left: 8px;
    border-top-color: #EC1C24;
    position: absolute;
    content: "";
    margin-top: 0px;
}

@keyframes slide-up {
    0% {
        opacity: 0;
        transform: translateY(20px);
    }
    100% {
        opacity: 1;
        transform: translateY(0);
    }
}
```
I have not study how to add them into jewel theme,
and have not enogugh knowledge of framework, like COMPILE::JS

These valiators are just a trial.
Please feel free to edit them or give me some hints, thanks


carlosrovira:
Hi Taro,
I'm reviewing your code. 
I'll try to integrate as well CSS
Thanks!

carlosrovira:
Hi Yu,
congratulations! your code is merged now! :)
I had to make some changes since it was not compiling just out of the box, and need to enter in namespaces xml to get available and recreate the example you did.
I setup de CSS in SASS temporarily. For example I think we don't have support for "keyframe" yet, so animation in that way is not working, but don't worry I'll make it work in other way as I get time.

Thanks! and keep the good work! :)

carlosrovira:
One more thing! For future contributions.

I think you started this code in other custom project, I recommend you code in Jewel library directly (in case is a change for Jewel), and add the configuation you need (check my commit to see the changes I did so you can do this in the future)
In this way your changes can be forwarded without more reviewer changes to repo.
If you continue this work we can propose give you committer access, but to get this privileges, you must to master this little things in future commits, so others will see that you can go almost alone.
Thanks! :)

carlosrovira:
Hi Taro,
I forget to add in the example, in Form tag, the attribute action="doSubmit(event)", like I see in the screenshot of the code. But seems the attribute "action" doesn't exists. Maybe you forgot to include in your PR?
thanks

carlosrovira:
Hi Yu,

I think I finish to accommodate your Validators in Jewel, integrate the styles in the themes and refactor to its own package. So mostly, things are now set in a right location in Jewel. As well refactored example to have its own section in JewelExample.

I see only two issues that hope you could have a look:

1.- in IE11 tooltip doesn't place in the right position. I saw that top and left styles are not setting right in IE11, it seems a problem of get the right javascript applies, maybe some instruction used to set top and left in IE11 is not recognized well by this browser.

2.- the "action" attribute in Form like I reported before is missing

Hope you like the integration of your code in Jewel.

Great work! :)

yestaro:
Hi Carlos
Thank you for merging the code.

1.  I will check IE11 problem.  Because I am not sure if the Validator is suitable for Jewel, so I did not full test, only test it on Chrome,
1.  the "action" attribute, I rename it to "valid" attribute. 
  In my first version, the FormValidator behavior as following
- when succeeds, will dispatch "action" event
- when fails, if user has set FormValidator.requiredFieldError, will alert the message.

But then I change FormValidator behavior
- when successed, dispatch "valid" event  
- when fails, dispatch "invalid" event
<Form action="..." is as web developer knows, but "valid" / "invalid" event is more like old flex.

I think I will modify the code again, back to "action" event
- when successed, dispatch "action" event  
- when fails, dispatch "invalid" event, and if user has set FormValidator.requiredFieldError, alert the message.

carlosrovira:
Hi Yu,

1.- I tested in Chrome, FF, Safari and IE11. Only IE11 give problems. In this case the good notice is that is not a CSS problem, but a JS problem. If you don't catch it I'll can take a look.

2.- I think VALID/INVALID is more consistent than ACTION/INVALID. Think that Royale try to abstract browsers, flash plugin and some day native, so we don't need to conform to some concrete platform things (in this case browsers, with action).
I think this:

> when fails, dispatch "invalid" event, and if user has set FormValidator.requiredFieldError, alert the message.

would be great addition.

thanks! :)

yestaro:
Hi Carlos

1.  CSS style:top..left... not appear in IE11,  as you say, it's JS problem. .... I have no idea about that. please take a look, thanks
2.  Thanks for your advice, I'll do it later.

PR #295, opened 2018-09-14
changes on 14-sep-2018

pashminakazi:


aharui:
This PR looks like maintainAspectRatio got added to UIComponent.  I don't think it should go there.  You can check the index in the ASDoc for Flex to see where it was originally.

PR #304, opened 2018-09-27
NumericStepper -  enter value into textinput should dispatch ValueChangeEvent

yestaro:


carlosrovira:
Hi Yu, I need a bit of background to understand this change. Currently NS dispatch a valueChange event as the first example in NumericStepperPlayGround shows. why we need to add an additional value change event?. One thing we should do is separate concerns and create a controller for this component that still is missing. Thanks for your thoughts on this.

yestaro:
Hi Carlos, If key in value on TextInput directly (not click spinner), no event triggered.
I think user enter the TextInput value directly should dispatch valueChange event.
In ListPlayGround that will be useful, user can enter the value in TextInput and goto the item quickly.

carlosrovira:
Ok Yu, thanks for the explanation, didn't have that in mind. Change is already merged. Thanks for this new contribution! :)

PR #313, opened 2018-10-09
emulation components add and/or modified

nihavend:


aharui:
Looks good.  It looked like at least one file had a [ResourceBundle] metadata that wasn't commented out.  It would be best to comment them out until we get ResourceBundle working somehow in JS.

If you can take a minute to check for and comment out [ResourceBundle] usage then we should be ok to accept this.

nihavend:
Yes ı can comment. I checked for the compile errors  but all the sources compile without error.




Is it neccessary to remove RB ?




Android için Outlook uygulamasını edinin







On Tue, Oct 9, 2018 at 7:10 PM +0300, "aharui" <notifications@github.com> wrote:












Looks good.  It looked like at least one file had a [ResourceBundle] metadata that wasn't commented out.  It would be best to comment them out until we get ResourceBundle working somehow in JS.


If you can take a minute to check for and comment out [ResourceBundle] usage then we should be ok to accept this.

aharui:
The SWC will compile, but the application that uses those classes will report an error about missing bundles right now, so we are commenting out all RB metadata for now.

If you can comment out the RB metadata, it will also help ensure that your pull-request workflow is set up correctly.  We had some other early contributors have trouble with updating a PR.

nihavend:
Completed @aharui. You can check it.

aharui:
Thanks and congratulations!

PR #314, opened 2018-10-09
ComplexSortField.as moved to the right package.

nihavend:
@aharui one of the classes (ComplexSortField.as)  was in wrong path, i moved to the right place. You can check the PR flow with this one maybe.


aharui:
Good catch!  Thanks!

nihavend:
Thanks.

PR #316, opened 2018-10-13
Emulation components created mostly related with spark DataGrid

nihavend:
Hi,

It is ready to be merged @aharui


aharui:
Do you really need DownloadProgressBar and the Touch-related classes?  Otherwise, try commenting out references to them.

nihavend:
While loading some modules there is a need for progress bar. 
Here : https://github.com/likyateknoloji/pinaraui-royale/blob/a145aac0a1b479cda48c5f5c673cd6d2a996dd4f/PinaraUI/src/com/likya/pinara/utils/DownloadProgressBarSubClassMin.as
For touch-related classes, i did not think so much on the subject. Which classe you want to be commented out ?

likyateknoloji:
If you say that there is not going to be the equivalent of any type of indicator to show that the application is being downloaded, than i may remove the emulation and dependency for sure. 
While working on emulation, i am not 100% percent sure about the components to be emulated. I am trying to not to go so deep on dependencies and conveting the simple implementations.
I am going to work on the issue in the evening or at least tomorrow.

nihavend:
If you say that there is not going to be the equivalent of any type of indicator to show that the application is being downloaded, than i may remove the emulation and dependency for sure.
While working on emulation, i am not 100% percent sure about the components to be emulated. I am trying to not to go so deep on dependencies and converting the simple implementations.
I am going to work on the issue in the evening or at least tomorrow.

PR #317, opened 2018-10-16
Service classes emulated

nihavend:
Http Service related classes emulated


aharui:
It looks pretty good, but there are some places where a method has both a COMPILE::SWF version and a COMPILE::JS version and there aren't any obvious differences between them.  It would be better to have one multi-platform version without COMPILE::SWF/COMPILE::JS and put COMPILE::SWF and COMPILE::JS blocks in the method body for any differences.

In some cases, the fact that there are different imports for JS vs SWF is good enough.  The same method body will use the JS import when compiled for JS and the SWF import when compiled for SWF.

Also, I don't see any need to try to retain any of the RESULT_FORMAT_XML code for the SWF version.  We should just leave it out until someone really needs it.

nihavend:
Thanks Aklex. I am going to check the classes for SWF/JS compiler directives. I commented all RESULT_FORMAT_XML as i remember. Where do you mean ?

nihavend:
I removed all  xml thins from sendBody method.

nihavend:
In DirectHTTPChannel:
connectTimeoutHandler are commented-out. Do you want me to enable ?

You shouldn't have to qualify flash.net.URLLoader vs org.apache.royale.net.URLLoader so methods that use it can be the same. 
I did not get the point.

nihavend:
AbstractOperation : Yes sure you are right. I made it single.

aharui:
In DirectHTTPChannel, you can leave connectTimeoutHandler commented out, but I think there only needs to be one of them, not two

frameworks/projects/MXRoyale/src/main/royale/mx/messaging/channels/DirectHTTPChannel.as
```diff
@@ -0,0 +1,675 @@
...
+		 *  @playerversion Flash 9
+		 *  @playerversion AIR 1.1
+		 *  @productversion BlazeDS 4
+		 *  @productversion LCDS 3 
+		 */
+		COMPILE::SWF
+		override protected function getMessageResponder(agent:MessageAgent, 
+														message:IMessage):MessageResponder
+		{
+			return new DirectHTTPMessageResponder(agent, message, this, new flash.net.URLLoader());
```
aharui: I think this can be:

return new DirectHTTPMessageResponder(agent, message, this, new URLLoader());

And then you don't need a separate version for COMPILE::JS

nihavend:
@aharui i think it is ok now.

PR #321, opened 2018-10-18
HttpService and related classes emulated

nihavend:
@aharui may you merge pls.


nihavend:
Hi @aharui, may you please merge my pull request.

frameworks/projects/MXRoyale/src/main/royale/MXRoyaleClasses.as
```diff
@@ -68,6 +68,7 @@ internal class MXRoyaleClasses
 	import mx.effects.Tween; Tween;
 	import mx.system.ApplicationDomain; ApplicationDomain;
 	import mx.rpc.http.HTTPService; mx.rpc.http.HTTPService;
+	import mx.rpc.http.mxml.HTTPService; mx.rpc.http.HTTPService;
```
aharui: This line shouldn't be necessary.  The mx-royale-manifest will take care of it.

aharui:
Do we really need IRepeaterClient?  Also NameUtil.  Would be nice to not have to support them in the future.

nihavend:
> Do we really need IRepeaterClient? Also NameUtil. Would be nice to not have to support them in the future.

I am not sure. Whan i was working on DefaultGridItemRenderer, i needed both and emulated. Than i gave for DefaultGridItemRenderer but these classes were ready and i included them in PR. I can remove if you say so.

frameworks/projects/MXRoyale/src/main/royale/MXRoyaleClasses.as
nihavend: Okay, i had missed adding mx-royale-manifest to compiler arguments.

nihavend: Removed.

aharui:
We should try not to include code we don't want to support forever.  I'm not sure NameUtil would be critical to anyone's application.  Repeater may not be easily implementable on other platforms/languages.  So let's remove them.

PR #326, opened 2018-10-26
Emulated classes added.

nihavend:


aharui:
I sure hope we don't need DescribeType*.  It is very Flash-specific.  What code needed it?

PR #335, opened 2018-10-29
Ajwfrost/externalinterface

ajwfrost:
Adding the implementation of ExternalInterface into the mx.external package - see other discussion within PR #325


aharui:
Is there a reason you chose to create a DIV as opposed to some non-visual element like a Script or something?  FWIW, we use lowercase names in createElement.

ajwfrost:
No reason really for DIV other than I'd seen it done before elsewhere.. presumably this could change to an empty Script element and would still work - I'll update it and give it a quick test...

aharui:
This is optional, but it occurred to me that if you add an "id" property that sets the HTMLElement's id property, and document that it only works for JS, that existing code that used document.getElementById() with some other ID would work without modification.

PR #339, opened 2018-11-08
Spark Module class added.

nihavend:


aharui:
Doesn't spark-royale-manifest.xml need to change to point to Spark Module?  I think it currently points to MX Module via lookupOnly

nihavend:
I am using spark spark.modules.Module so added mapping. Anyone using mx.modules.Module may stilll need that mapping i guess. If not than i canre remove mx version.

nihavend:
It is more clear now, I am not sure that spark Module is working as mx Module. I am not at the point of things to check if working but trying to remove compiling errors.  It is better to leave as it is and remove psark mapping i guess.

And my question, how can i get rid of compiler arros for s:Module in my code then ?

aharui:
It depends on what errors you are getting.  If you are just referencing the class from AS, you can include the class in SparkRoyaleClasses.as but not change it in the manifest.  If you are referencing it from MXML, I'm curious as to what APIs are missing.

aharui:
If you are using Spark Module in MXML files, what errors are you currently getting if Module is mapped to MX Module?  Adding Spark Module to SparkRoyaleClasses won't help the Module tag be recognized as Spark Module.

nihavend:
I should have been more clear @aharui. Spark Module is part of MXML but in the action script part of mxml. 

So here are the steps and the error :

1. comment out the import in SparkRoyaleClasses
2. build spark royale
3. build my project
4. get the error : 

```
D:\dev\royale-emulation-works\github\pinaraui-royale\PinaraUI\src\com\likya\pinara\comps\jobdetail\JobBaseInfo.mxml(19): col: 11 Warning: Definition spark.modules.Module could not be found.
			import spark.modules.Module;
			       ^
D:\dev\royale-emulation-works\github\pinaraui-royale\PinaraUI\src\com\likya\pinara\comps\jobdetail\JobBaseInfo.mxml(21): col: 32 Error: Type was not found or was not a compile-time constant: Module.
			private var remoteSchModule:Module;
			                            ^
```
Here is the part of script :
```
			import spark.modules.Module;
			private var remoteSchModule:Module;
```
Ref : https://github.com/likyateknoloji/pinaraui-royale/blob/master/PinaraUI/src/com/likya/pinara/comps/jobdetail/JobBaseInfo.mxml

aharui:
So doesn't adding the import to SparkRoyaleClasses (but not changing spark-royale-manifest.xml) make your errors go away?

I noticed that ContainerCreationPolicy is now in this PR.  Why was it added?  Hopefully we don't need it.

nihavend:
Exactly this is what i am saying @aharui. I added the import to  SparkRoyaleClasses and errors go away.

ContainerCreationPolicy was weak reference in Module, i removed it from MXRoyaleClasses and updated the PR.

PR #340, opened 2018-11-09
URLLoader: HTTPStatusEvent addition

hcsuk:
Adding an HTTPStatusEvent into the JS-based code. We've found that this is required by some applications and it seems to make sense to expose this to client code. Initial thought was to add a status property to the URLLoader directly, but the "httpStatus" event was already being dispatched, so this just adds an extra value to the event object so that listeners can check whether the request was successful.

### DIFF
--- a/frameworks/projects/Network/src/main/royale/NetworkClasses.as
+++ b/frameworks/projects/Network/src/main/royale/NetworkClasses.as
@@ -32,6 +32,10 @@ package
         import org.apache.royale.net.URLBinaryUploader; URLBinaryUploader;
         import org.apache.royale.net.events.ResultEvent; ResultEvent;
         import org.apache.royale.net.events.FaultEvent; FaultEvent;
+        COMPILE::JS
+        {
+            import org.apache.royale.net.events.HTTPStatusEvent; HTTPStatusEvent;
+        }
         
         import org.apache.royale.net.remoting.messages.AcknowledgeMessage; AcknowledgeMessage;
         import org.apache.royale.net.remoting.messages.AcknowledgeMessageExt;
--- a/frameworks/projects/Network/src/main/royale/org/apache/royale/net/URLLoader.as
+++ b/frameworks/projects/Network/src/main/royale/org/apache/royale/net/URLLoader.as
@@ -27,6 +27,10 @@ package org.apache.royale.net
         import flash.net.URLRequest;
         import flash.net.URLRequestHeader;
     }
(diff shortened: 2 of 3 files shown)


aharui:
I'm accepting this change, but I want to bring up a pattern change between Flex and Royale.  IMO, the PAYG Royale components should have as few Event subclasses as possible.  Each new class definition adds to the amount of code that needs to be downloaded and parsed at startup.  So, if you look at how Royale currently handles Timer, there is no TimerEvent, it just sends an Event.  The constant that represents the name/type string of the Event is defined in the Timer class itself because you are already paying for that class definition.  Ideally, Royale would have Event and ValueEvent (which allows an Event to have a payload of one data item, and code needing to dispatch events would just dispatch those events and not subclasses of those events.  There is a ValueChangeEvent that has two fields (before and after values).  There could be one where the value is known to be a String if that's helpful.  Unless you actually have code in the listener to check the type, in JS, the type will not actually be checked (it is checked by the verifier in Flash, though), so what type is actually sent generally won't matter and it is much faster to check the name/type string than run Language.is to check the class type.

hcsuk:
Thanks - I had wondered about using the ValueEvent .. my main concern with that was the typing, where it's after an Object type. If it was a * it might have been possible I guess, but when I wanted to put an integer into this slot I would have got compiler errors from the initial type checking. Adding it as an object with a 'status' member would have added overhead within the VM. As an aside, this might be where a templated type would be useful e.g. `ValueEvent.<uint>` ... but that's a bit of a digression!
cheers

aharui:
We could change ValueEvent to use * instead ofObject.
We could add a ValueIntEvent or ValueEventInt if retaining strong-typing is important.  The main goal is to reduce the number of classes that get loaded and parsed at startup that have little usefulness at runtime.

hcsuk:
Having the * type makes me a little uncomfortable but that's just because of my background in strongly typed languages probably! and also knowing what happens under the hood within AS3/avmplus around objects and references. But I definitely agree with the idea of reducing the number of classes required so would probably go with this option rather than creating multiple ValueEvent derivations for the different types..

Happy to change to that, and then revert these changes, if you think this is worth doing?

thanks

aharui:
IMO, Object is as weak as *.  Object is used in Flash often as "Any but not undefined".  So changing to * probably wouldn't make things significantly weaker.

If you have time to make these changes, please do so.  I think having fewer Event classes will benefit Royale.

PR #341, opened 2018-11-09
Adding listener for when a bindable style has changed

hcsuk:
There was already an event handler for this, but it wasn't being hooked up... If we have some bindable styles, they are being updated by the binding code but the style wasn't being re-applied to the element. Adding this event listener (if the style implements IEventDispatcher - the bindable one does, the simple one doesn't) made it all just work..

### DIFF
--- a/frameworks/projects/Basic/src/main/royale/org/apache/royale/core/UIBase.as
+++ b/frameworks/projects/Basic/src/main/royale/org/apache/royale/core/UIBase.as
@@ -1416,6 +1416,10 @@ package org.apache.royale.core
                 
                 if (style)
                     ValuesManager.valuesImpl.applyStyles(this, style);
+
+                // add an event listener if the style dispatches events (e.g. BindableCSSStyles)
+                var styleAsEventDispatcher : IEventDispatcher = style as IEventDispatcher;
+                if (styleAsEventDispatcher) styleAsEventDispatcher.addEventListener(ValueChangeEvent.VALUE_CHANGE, styleChangeHandler);
             }
             
 			if (isNaN(_explicitWidth) && isNaN(_percentWidth)) 


aharui:
Actually, I was thinking of pulling out that UIBase.styleChangeHandler.  Ideally, because of PAYG, we could devise a solution where BindableCSSStyles just runs the code that was in styleChangeHandler.

This is another example of the mindset-shift difference between Flex and Royale that I'm trying to promote.  Flex's UIComponent grew to 13,000 lines of code because we kept adding little bits of just-in-case code.  I would like us to try to find solutions that are PAYG and keep separation of concerns.

hcsuk:
Okay yes good point (and I definitely like the way that most of the Royale classes are very small - makes them easy to understand!). So I'll have a quick go at updating the `BindableCSSStyles` class - or perhaps, having looked at it, it would make more sense provide a subclass of `SimpleCSSValuesImpl` which is where there's an association between the UI element and the Style object (in `applyStyles`) where we could listen out for changes in the style, and then just directly adjust that particular style rather than having to create a new dynamic object and then parse it...

aharui:
It would be PAYG to have a variant of SimpleCSSValuesImpl that expects runtime changes to styles, but couldn't all of the code that needs to be run be encapsulated in BindableCSSStyles (and avoid the temporary object?  That would be even more PAYG, if the code doesn't get awful or too much code gets duplicated from the IValuesImpl.  Or maybe applyStyles need to be refactored or a new API is needed on IValuesImpl to be passed the BindableCSSStyles itself.

hcsuk:
To do this, there would need to be a link created between the UI object (UIBase), and the BindableCSSStyles object, so that the latter could listen out for style changes on the UIBase object. So this would imply we still need to change UIBase.as, such that when the style is set on it, it sets up the event listener - essentially this just moves the `styleChangeHandler` code from UIBase into BindableCSSStyles, but still needs similar changes to be made to UIBase.

An alternative would be to create a bead that does this stuff so we could keep the same code in all the existing classes, but just have folk add a bead that would set everything up to sort out the styles based on the changes to the UIBase object... that would be a little more stand-alone and avoids adding code into any existing class or changing the current functionality for anyone..

aharui:
Yes, good point that BindableCSSStyles needs a reference to its "host".  

So thinking about this some more, there are two scenarios to support.  One is that someone sets a style property at runtime via 

`myUIBase.style.foo = "bar";`

And one just I thought about is that there could be properties on a component that are implemented as changes to HTMLElement.styles.  The 'visible' property is an example (more or less).  Not sure if there are others.  I would say that properties that result in changes to HTMLElement.styles are not required to go through BindableCSSStyles.  Instead they can just set HTMLElement.styles directly.

So, I'm not clear that BindableCSSStyles needs to listen to UIBase for style changes, but it does need a reference to know what to pass to the IValuesImpl if we encapsulate change handling outside of UIBase (which would be more PAYG, IMO).

If I understand your alternate plan, you are suggesting a bead that integrates BindableCSSStyles with its host.  That's reasonable, but isn't another option that BindableCSSStyles itself could a bead?  Not sure if we have other beads assigned via properties.  The styles setter would test if the object is IBead and if so, would put BindableCSSStyles on the strand.  That test is still just-in-case code, but it might be ok.  We could get rid of that test by requiring that if folks use BindableCSSStyles that they use it on the beads list and not as the styles object, but that seems too weird.  Or maybe all objects assigned to styles must be a Bead (or a String).

Yet another option could be that the styles setter tests for an interface and assigns the reference.  There is already an IDocument interface, but just having a host reference is more like IStrand so maybe having BindableCSSStyles be a bead is better and avoids people forgetting to add the bead that ties BindableCSSStyles to the change handling code.

Ideally, we would remove stylesChangeHandler from UIBase and put it in BindableCSSStyles and folks could use BindableCSSStyles without any extra beads and everything would just work without too much other just-in-case code.

hcsuk:
Okay - I'm having a bit of a play with this and will see what works.. about to stop for the evening to become a taxi service for a bunch of kids..!

aharui:
Thanks for experimenting.  I agree with your sentiments about the amount of code in the styles setter.

Some options we have are to dictate that:
1) the styles property can only be set to an object once (you can set it to a string multiple times since you can do that in HTML).  Then you wouldn't need to remove the bead from the strand
2) require that all objects (except Strings) assigned to the styles property be IBeads.

In addition, it might be that we can alter the API of applyStyles to take a third parameter that is a list of properties to set.  Then you wouldn't have to create the temporary object and can just pass in the entire BindableCSSValuesImpl.

That might cut down on the amount of code in UIBase "just-in-case".

hcsuk:
One question: if you set the "style" property via a string, and make multiple calls, should the result be incremental? e.g. `myObj.style = "backgroundColor: blue";` followed by `myObj.style = "backgroundAlpha: 0.5";` - would we end up with an object that still has the blue backround now at 50% opacity? I'm assuming so .. which means that the current code is a bit inefficient, it will keep creating an object that's then used in the `applyStyles` method.

So if a string is passed in, we could either just apply this directly within `applyStyles` without caching the resulting object as the `_style` member variable, or alternatively we could build up a `_style` object that is extended each time the call is made. I'm not sure whether it's worth doing this though - given our proposed paradigm would mean that we then need to ensure that the dynamic object that we create implements the `IStyleObject` interface.

So I'm thinking that setting a style by string just directly applies that style but doesn't create the object; is this okay i.e. no one wants to read back the style afterwards? And when setting a style by object, it needs to implement the IStyleObject interface and then in the Bindable version we can add the necessary logic to ensure any updates to it are applied to the element..

aharui:
The String is supposed to work like the HTMLElement.style in HTML markup.  I don't think it is "incremental", it defines the total set of local styles in the CSS lookup.  Setting a SImpleCSSStyles or BindableCSSStyles does the same thing.

If don't think you can read back the string you set in HTML markup, so I would say it is fine for us to convert it to an object.

hcsuk:
Okay thanks; so I've looked into more what's happening in various places with this... I'd wondered whether there would be value in defining a base class for the various Style objects that we have, which would implement the IStyleObject; this would then be an object that some of the client code (e.g. in mx.controls.VRule) could use plus the `parseStyles` functions in SimpleCSSValuesImpl/AllCSSValuesImpl, which are currently just returning an untyped object. So this then ensures we have a style object of known type, and can set it's "host object"; in the Bindable versions, this is what we then cache in order to apply the styles to the host object when one of our styles is changed.

The other thing I would propose is to shift AllCSSValuesImpl so that it derives from SimpleCSSValuesImpl. They're pretty much identical apart from the lists of style categories that are used within `applyStyles`, and it looks like there has been some maintenance/bugfixing done on one and not the other, so it could be a maintenance problem. I'd actually wondered whether to shift the `parseStyles` into one of the Style objects (i.e. we have an implementation `SimpleCSSStyles.parseFromString()` and an `AllCSSStyles.parseFromString()`) but there still needs to be a way for the user to define the list of styles that are supported and the ValuesImpl object is a global way of doing that.

I'll translate the above comments into code and push them here so you can see what I mean...

aharui:
If you have time to refactor things so more code is shared that would be great.  There are plenty of places that could stand to be refactored.

aharui:
One thing to keep in mind is to be careful about putting a lot of infrastructure in the base classes in order to make the subclasses more efficient.  Because of PAYG, we will lean towards having the subclass do more work so the base class isn't carrying too much code just in case there is a subclass.

hcsuk:
Yes I know what you mean re. adding too much in base classes.. 

Just on the `SimpleCSSValuesImpl` vs `AllCSSValuesImpl` case: there aren't actually any significant differences between these classes; lots of minor diffs which are due to people updating one class with small changes but not the other (mostly updating the 'simple' one e.g. `if (o)` had been changed to `if (o !== undefined)` etc. Then the only real functional changes are:

1. AllCSSValuesImpl has two additional properties being set on a 'conditionalCombiners' object
2. AllCSSValuesImpl would avoid putting "px" on the end of a fontWeight style
3. SimpleCSSValuesImpl handles style strings passed in as "some-thing" and changes them to "someThing"
4. SimpleCSSValuesImpl has an additional code block to implement `addRule` in JavaScript.

So I can't really see a good reason to not just combine these two; the additions that are in 'simple' I definitely think should be made in 'all' as well.. I guess we could have those extra changes from 'all' kept out of 'simple', but I would have thought the argument for that was slim.

From the description, I had expected there to be a whole extra set of styles that the 'all' class supported, but it looks to me like these classes don't really have so much of the knowledge of the actual styles - unless it's around the various categories i.e. which ones are colors, which are numbers, etc.  The initial commit comment suggests the creation of the 'all' class was to slim down the simple case, so perhaps we still want the 'all' class which would be where future style categories could be added; but 99% of the functionality would still be common so inheriting this from the 'simple' case seems a good idea to me..

aharui:
IIRC, the differences are around how to handle the "categories".  The Simple version is really intended to be closer to what was available in Flex and easily implementable in SWF.  So "fontWeight" is expecting only "bold" and "normal" and not numbers.  The All version should handle numbers for fontWeight, and eventually learn all kinds of things like various units ("em", etc).  It doesn't right now because we didn't have time.  You are not obligated to deal with that now.  Lots of things in Royale are somewhat incomplete and get done when someone actually needs them.

Some fixes to Simple like handling '-' should work in All.  Hopefully by subclassing, that will happen.

So, IMO, there really should be a Simple and a more powerful All and I would be careful about making Simple carry around code just in case it is being used as the base for All.

hcsuk:
So I've just put in a few changes to Simple and now the All one can derive from this and will be able to add in different style categories if needed in the future - fontWeight is in there for now, just to ensure we keep the functionality the same.

Will look at the UIBase 'style' setter now and try to keep it simple..

hcsuk:
Given that we have some dynamic objects being created for styles (e.g. when parsed from a string), and also given the simple style types don't need a reference back to the host UI object, I'm thinking that it would add unnecessary complexity to force our 'style' objects to all implement a particular interface - it doesn't seem worth creating a new base type just for this, when most of the time it won't actually be needed.

So I'm thinking perhaps it's just the bindable styles that need to have the host object to be set through the `IStyleObject` interface (possibly renaming this to `IBindableStyleObject` for clarity?). Given they're already deriving from EventDispatcher, and there are only two bindable style objects currently, we can just add the implementation/listener code directly into them. This means the 'simple' case stays as simple as it currently is, and we're only impacting the 'bindable' ones to make them work properly..

I'll make the updates shortly so that you can see what I mean via code..

PR #345, opened 2018-11-15
Adding support for posting data

hcsuk:
Taking the information already present in the URLRequest object and setting it appropriately on the XMLHttpRequest within the URLLoader object

This then allows us to send a 'post' request with strings or with binarydata objects on a JavaScript build (it would already work on a SWF build due to the use of the flash.net.URLLoader object).

### DIFF
--- a/frameworks/projects/Network/src/main/royale/org/apache/royale/net/URLLoader.as
+++ b/frameworks/projects/Network/src/main/royale/org/apache/royale/net/URLLoader.as
@@ -192,21 +192,28 @@ package org.apache.royale.net
                     }
                 }
                 
-                /*
-                if (request.method != HTTPConstants.GET &&
-                    !sawContentType && contentData) {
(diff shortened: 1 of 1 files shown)


Harbs:
I'm pretty sure I've bumped into a case where setting content type headers broke APIs. I don't remember the details.

FWIW, I've been using URLBinaryLoader pretty exclusively. If I were to work on URLLoader (probably a good idea), I'd model it after the way that works. (i.e. offload the implementation to URLStream)

Harbs:
I'm also pretty sure the idea behind URLLoader was that it would not have a dependency in BinaryData. @aharui Can you comment?

If a dependency on BinaryData is OK, then modeling it after URLBinaryLoader might be the way to go.

Harbs:
Added reference in doc 3ac7946621b65bc6f92afe9936cccad543845723

PR #348, opened 2018-11-17
Updating ExternalInterface to cope with qnamed functions

hcsuk:
If a function name of 'window.object.fnc' is passed in, we now correctly parse this string to get a reference to the 'fnc' and call it with the 'window.object' as the 'this' parameter


Harbs:
Is this PR still relevant? We should either fix and apply or close.

greg-dove:
Closing this PR as it was already implemented (in isolation) in a similar way. Additional compatibility updates will be forthcoming (i.e. support for arbitrary script injection, which is also supported by flash ExternalInterface).

PR #349, opened 2018-11-20
Added status variable and getter for JS compiler.

nihavend:
Added status variable and getter for JS compiler.

### DIFF
--- a/frameworks/projects/MXRoyale/src/main/royale/mx/events/HTTPStatusEvent.as
+++ b/frameworks/projects/MXRoyale/src/main/royale/mx/events/HTTPStatusEvent.as
@@ -19,91 +19,91 @@
 
 package mx.events
 {
-COMPILE::SWF
-{
-import flash.events.HTTPStatusEvent;        
-}
-/* import mx.events.Event;*/
-import org.apache.royale.events.Event;
-import org.apache.royale.events.IRoyaleEvent;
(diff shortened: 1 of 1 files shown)


nihavend:
@piotrzarzycki21  and @aharui do you have any concerns ?

frameworks/projects/MXRoyale/src/main/royale/mx/events/HTTPStatusEvent.as
```diff
@@ -19,91 +19,95 @@
...
+		 *  @langversion 3.0
+		 *  @playerversion Flash 9
+		 *  @playerversion AIR 1.1
+		 *  @productversion Royale 0.9.3
+		 */
+		public function HTTPStatusEvent(type:String, bubbles:Boolean = false,
+										cancelable:Boolean = false, status:int = 0)
+		{
+			super(type, bubbles, cancelable);
+			this.m_status = status;
```
aharui: I don't think you need 'this.'

There should be very few occurrences of 'this' in the AS source.  I think should only be needed when referencing member APIs that have the same name as local and parameter variables.

nihavend: Completely agree. They are irrelevant. Removed.

PR #357, opened 2018-12-25
Nested includeIn and/or excludeFrom

likyateknoloji:
Using nested states (includeIn or excludeFrom) causes parent to be null in some cases, added null check to prevent exception.


nihavend:
Unfortunately, I am not able to create an isolated test case, but ,if you want I may share you my project, logs etc. which can prove that there is "NullPointerException" when trying to get parent with nested state usage.

Harbs:
Thanks for the email.

You're probably right that the null check s the way to go on this (at least for now). If I understand correctly, what's happening is the implementation is trying to add to an (parent) element which is disconnected due to it being in a non-included state.

PR #406, opened 2019-03-24
Related with URLLoader emulation, HttpStatus event creation and request parameter type checking.

nihavend:
URLLoader class is added to emulation
Generic event checked and converted to HttpStatusEvent
Parameter type is checked and the extracted string value with the appropriate method

@aharui may you please check the PR


nihavend:
@aharui kindly remind

PR #410, opened 2019-04-04
fix removeElement if child has 'other' positioner

yestaro:
EX: Jewel TextInput has a positioner(div) and element(input)
when jewelCard.removeElement(jewelInput) 
Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at org.apache.royale.jewel.Card.org.apache.royale.core.UIBase.removeElement


carlosrovira:
Hi Taro,

the PR is just this part: "fix removeElement if child has 'other' positioner" right?
this was included by mistake: "Merge pull request #2 from apache/develop" ?
thanks

@aharui : I'm testing to remove a Jewel Textinput and I found this issue. Do you think is ok that in UIBase line 1380

`element.removeChild(c.element as HTMLElement);`

should be:

`element.removeChild(c.positioner as HTMLElement);`

thanks

aharui:
Using positioner instead of element makes sense.

yestaro:
Hi carlos
Yes, just this part: "fix removeElement if child has 'other' positioner

carlosrovira:
Hi 
@aharui thanks for let us know :)

I push the fix and give credits to @yestaro, not use this PR since seems malformed in part

@yestaro, I think I saw more code using c.element that maybe could be c.positioner. Could you check it and provide more fixes like this if you find it? Thanks for submitting it! :)

yestaro:
Hi 
@carlosrovira 
I have checked UIBase.as, I think only getElementIndex() need to use positioner instead of element.
and created new pull request, thanks

PR #484, opened 2019-09-27
fix Channelset login/logout throwing an Error

mrvx23:
Found a bug with RemoteObject login and logout (partly) failing. The issue is revealed when doing a logout and re-login on an already authenticated channel.

Root cause:
internalSetCredentials is in mx_internal namespace and the name does not correctly get converted in javascript transpiled code.

Fix:
added explicit cast of array element to base class for compiler to pick up correct function naming.


greg-dove:
That looks good to me, and very clear description... thanks!

Issue #486, opened 2019-09-30
stateChangeComplete event is fired if set again same state (bug or not) ?


cristallium:
Hi,

Suppose you have 2 states : **A** and **B** and currentState is **A**
Then setting currentState again to **A** will fire _stateChangeComplete_.

I don't know if we can consider this as a bug or not... On Flex component setting twice the same state doesn't fire _stateChangeComplete_

There is a simple workaround : just test the currentState before setting it.

Regards

carlosrovira:
Hi, this seems to me a bug. change event should only be dispatched if state is not the same.

cristallium:
Merged https://github.com/apache/royale-asjs/commit/42cd0c7c450f385a8fb2e5194cb1c7660578cd6b

PR #492, opened 2019-10-03
Update issue#486 (part1) prevent currentStateChange event to be fired if set same state as current

cristallium:
Don't dispatch "currentStateChange" event  if setted state is the same as _currentState (see issue #486)

### DIFF
--- a/frameworks/projects/Basic/src/main/royale/org/apache/royale/core/GroupBase.as
+++ b/frameworks/projects/Basic/src/main/royale/org/apache/royale/core/GroupBase.as
@@ -254,6 +254,7 @@ package org.apache.royale.core
          */
         public function set currentState(value:String):void
         {
+			if (value == _currentState) return;
             var event:ValueChangeEvent = new ValueChangeEvent("currentStateChange", false, false, _currentState, value)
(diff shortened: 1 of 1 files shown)


cristallium:
Hi Harbs,
As it's first time I'm using push request I have some difficult to use it.
I wanted to do only one request for prevent currentState to dispatch if set same state but I don't know how to do to I have done several patch.
Please note that for file frameworks/projects/Basic/src/main/royale/org/apache/royale/html/MXMLBeadView.as I have done a mistake : I commited directly on develop branch.

And also note that it seem to be missing a ";" the line just after my commits

To anwser your question "Anyone have a style preference on if(false)return vs. if(true{//do stuff}?" personnaly I prefer if (false) return because more less code and more readable code but some people doesn't want a return in a middle of function ... So you can do what you think the best

cristallium:
>Please note that for file frameworks/projects/Basic/src/main/royale/org/apache/royale/html/MXMLBeadView.as I have done a mistake : I commited directly on develop branch.
Well in fact I struggle with commit, I have done  it on my fork. I have 3 more files not commited. I search how to do it ...

PR #493, opened 2019-10-03
Update issue#486 (part2) prevent currentStateChange event to be fired if set same state as current

cristallium:
Don't dispatch "currentStateChange" event  if set state is the same as _currentState (see issue #486)


cristallium:
Please could you check ";" at end of line just after my line insertion

Harbs:
It looks like the semicolon is missing. Feel free to add it.

PR #576, opened 2019-11-19
ExternalInterface.call not correctly resolving object function calls

mrvx23:
Issue occurs if attempting to call a sub-function of an object via ExternalInterface.call (e.g. BrowserHistory.getUrl)
ExternalInterface attempts to find this method via direct string search on window, e.g. window["BrowserHistory.getUrl"]. This fails, because BrowserHistory is an object, with getUrl function.

Fix:
Added recursive resolution of all function name parts to arrive at the correct function reference to invoke. 
This allows ExternalInterface to correctly look up the function to call, e.g. window["BrowserHistory"]["getUrl"].


mrvx23:
Duplicates #348, which looks to be coded in a more elegant manner, taking into account the parent object reference when invoking the function. Please choose one of the two for merge, as the bug is otherwise affecting usage of the SDK.

mrvx23:
The commit for fix of #580 also got included here by accident.

mrvx23:
Closing, as it duplicates #348.

PR #580, opened 2019-11-22
Fix navigationlink href value always being #

mrvx23:
NavigationLinkItemRender reads the href value of the link out of it's data object .href property.

Expected:
NavigationLinkItemRenderer gets compiled into an anchor tag with the specified href value.

Actual:
NavigationLinkItemRenderer gets compiled into an anchor tag with href="#".

This is due to the internal logic of processing the href value, which seems to have broken due to indirect changes in (parent?) components. Href is only considered if an internal textNode is present, but in reality, the anchor tag and the child textNode should be handled independently.

### DIFF
--- a/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/itemRenderers/NavigationLinkItemRenderer.as
+++ b/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/itemRenderers/NavigationLinkItemRenderer.as
@@ -139,11 +139,20 @@ package org.apache.royale.jewel.itemRenderers
 			}
 			// text = getLabelFromData(this, value);
 			
-            if(value.href !== undefined)
-			{
-				if (value.href === null) _href = "#";
-                else _href = String(value.href);
-			} else _href = "#";
+            if(value.href === undefined || value.href === null) 
+				_href = "#";
+			else {
+				// check if link is absolute
+				if (String(value.href).indexOf("http://") == 0 || String(value.href).indexOf("https://") == 0)
+					_href = String(value.href);
+				else {
+					// if we have a relative link, make sure it's prefixed with '#' to stay on the app page 
+					if (String(value.href).indexOf[0] != '#')
+						_href = '#' + String(value.href);
+					else
+						_href = String(value.href);
+				}
+			}
 
 			COMPILE::JS
 			{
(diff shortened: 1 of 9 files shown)


carlosrovira:
Hi this PR seems to be mixed and with duplicates. Can you please redo it so I can review it. 
Thanks!

piotrzarzycki21:
@mrvx23 Did you build TourDeJewel and check whether there is no changes in UI due to those changes ?

mrvx23:
@piotrzarzycki21 you're right, after checking I see that this breaks the navigation drawer in TourDeJewel.

The problem is that the 'href' property can possibly be used for two purposes:

1. To identify the navigation section for the change handler, so that the correct section can be made visible (= "internal link"), and
2. To act as a standard link/anchor reference, to support deep linking with URL in a Royale application.

Obviously 2) is currently not functional in develop branch, but at some point it used to be supported, as there is (dead) logic there to set the a.href to another value than '#'.

Solution proposal:
Inspect the incoming href value. Determine if it's an absolute or relative link. If it starts with http:// or https://, treat it as an absolute link and pass it through to the anchor tag.
Otherwise, treat it as a relative link, and inspect if it starts with prefix '#'. If it does not start, add the prefix '#'. This creates backwards compatibility for existing code (like TourDeJewel) that rely on href as an internal link.

mrvx23:
I committed also the proposed fix for backwards compatibility. Note that browser URL will still reflect the new fragment, but this will not break standard Navigation usage.

Please see the change under Files changed tab. The commit history got a little mixed up, but you will see it's only the single file that's changed. If you prefer, I can try creating a new pull request with a single commit.

piotrzarzycki21:
@mrvx23 Well I'm not a fun of checking whether some string has searched value or not. Maybe in that case this is the only solution. How my link should behave if user put's there "ftp://" ?

piotrzarzycki21:
> Hi this PR seems to be mixed and with duplicates. Can you please redo it so I can review it.
> Thanks!

What commits are mixed ? I see that this PR contains stuff which is not accepted yet from other PR (https://github.com/apache/royale-asjs/pull/576) - is this what you are referring to ?

mrvx23:
Right, and also case differences could through it off (e.g. HTTP://). Perhaps the most generic solution is a regex like https://stackoverflow.com/questions/10687099/how-to-test-if-a-url-string-is-absolute-or-relative/19709846#19709846

mrvx23:
Closing, as pull request has mixed commits.

carlosrovira:
Hi @mrvx23, I think this is solved now, since recently I removed `href` since as you said was never used after other changes some months ago. I removed it to make Router class work with Jewel. Thanks for closing it. Please let me know if you find any other issue.

mrvx23:
Great to hear it's fixed, thanks for the followup, @carlosrovira !

PR #699, opened 2020-01-29
MultiSelectionCollectionViewModel small change

tzivz:


Harbs:
@yishayw Yes. This is for a `CollectionView` which by definition is not an array.

PR #707, opened 2020-02-03
- Implemented a js-only distribution

chrisdutz:
Refactored the assembies to allow js-only distributions.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -41,6 +41,10 @@
     <playerglobalHome>libs/player</playerglobalHome>
     <targetPlayerMajorVersion>20</targetPlayerMajorVersion>
     <targetPlayerMinorVersion>0</targetPlayerMinorVersion>
+
+    <assembly.bin>bin.xml</assembly.bin>
+    <assembly.dir>dir.xml</assembly.dir>
+    <assembly.sdk-description>royale-sdk-description.xml</assembly.sdk-description>
   </properties>
 
   <modules>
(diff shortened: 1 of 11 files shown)


carlosrovira:
Hi Chris, I tested both with/without SWF and all works well except IDE SDK compiling with js-only version. option-with-swf continues to work in all cases. I'm going to merge since it the issue is just for the new js-only distribution when compiling from IDE. Thanks for doing it! :)

piotrzarzycki21:
@carlosrovira Could you please drop on the list email with exact steps how to run both swf/js distribution build? I know there were some instructions in the other threads, but summary would be awesome. Thanks! 

Thanks @chrisdutz !

carlosrovira:
Hi Piotr, I already updated the main wiki page with actual info :). All is here:

https://github.com/apache/royale-asjs/wiki/Build-Apache-Royale-with-Maven

with this change you now can avoid "option-with-swf" profile for a JS-only build (no SWF). I'll update wiki page as soon as we fix the JS-only compilation with IDE.

Please, let me know if you see some issue in that page in order to update that page

carlosrovira:
Just fixed the issue with IDEs. It was just a matter of copying frameworks/libs to distribution.
Thanks Chris again for all the work done to get this JS-only maven distribution SDK!! :)

piotrzarzycki21:
> Just fixed the issue with IDEs. It was just a matter of copying frameworks/libs to distribution.
> Thanks Chris again for all the work done to get this JS-only maven distribution SDK!! :)

I still didn't check any of that with Moonshine, unless you did that by saying IDEs ?

carlosrovira:
Hi Piotr,
no, just tested with VSCode. My point is that if VSCode recognize Maven distribution as SDK, and was able to compile TDJ with it successfully then Moonshine should do that too, right?

PR #801, opened 2020-04-09
jewel: fix backward grid layouts for desktop & widescreen

brianraymes:
I've noticed that the Grid Cell Layouts were reversed for Desktop and Widescreen in TourDeJewel and my local testing. Found this backwards assignment while hunting around.

### DIFF
--- a/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/layouts/GridCellLayout.as
+++ b/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/layouts/GridCellLayout.as
@@ -76,8 +76,8 @@ package org.apache.royale.jewel.beads.layouts
 
 			COMPILE::JS
 			{
-				setFractionForScreen(DESKTOP, _wideScreenNumerator, _wideScreenDenominator);
-				setFractionForScreen(WIDESCREEN, _desktopNumerator, _desktopDenominator);
+				setFractionForScreen(WIDESCREEN, _wideScreenNumerator, _wideScreenDenominator);
+				setFractionForScreen(DESKTOP, _desktopNumerator, _desktopDenominator);
 				setFractionForScreen(TABLET, _tabletNumerator, _tabletDenominator);
 				setFractionForScreen(PHONE, _phoneNumerator, _phoneDenominator);
 			}


brianraymes:
Also, a bit curious about the sizes declared for desktop VS widescreen:
```
public static const DESKTOP_BREAKPOINT:Number = 992;
public static const WIDESCREEN_BREAKPOINT:Number = 1200;
```

Desktop seems a bit.. small? Also, are these possible to override for an application?

carlosrovira:
Thanks Brian for starting to contribute PRs! Hope to get more coming ;)

carlosrovira:
Hi Brian, 

about the bigger is widescreen, do you see something wrong?

About overriding. Think that in Jewel is really JewelTheme SASS framework where we define variables (check "_variables.sass" to see device dimensions. So overriding should required override in SASS. Jewel lib AS constants are just convenient but loosely coupled to the SASS theme, Maybe Responsive Sizes could be part of the JewelTheme instead of Jewel. Can take a look as I have some time

carlosrovira:
Hi Brian,

was checking about moving ResponsiveSizes class that holds all responsive constants to JewelTheme and is not a solution.

While a theme can have AS3 Views (as LabelViewBead.mxml example in that library), and that has sense since you can then wire it though CSS without make Jewel has any reference to that View Class, for AS constant files doesn't has sense, since you need to refer to it from code. So that defeats the target.

So maybe the only solution to make it as configurable as possible from user perspective could be use "js-define" [1] variables, so users can configure it when building the App.

Thoughts are welcome. This is not priority, so we can think more about it.

[1] https://apache.github.io/royale-docs/compiler/compiler-options.html#js-define

brianraymes:
> about the bigger is widescreen, do you see something wrong?

I wouldn't classify it as wrong, but I wonder if there are not enough sizes in general. With the increasing sizes of monitors and resolutions these days, I think it would be nice to be able to support more scaling options for larger screens.

I think what is throwing me off is the definition of widescreen a bit ambiguous and IMO, considerably larger than 1200px (WUXGA).

I'm wondering if it would make sense to include options beyond Tablet/Desktop such as UWHD, WQHD, UWQHD, QHD, 4K, etc.

> So maybe the only solution to make it as configurable as possible from user perspective could be use "js-define" [1] variables, so users can configure it when building the App.

I was mostly curious if these breakpoints could be modified per app as I wasn't sure how they are referenced. It sounds like it would be quite difficult unless we added more sizes in general.

carlosrovira:
Hi Brian,

more sizes are welcome to bring more PRs with more sizes (maybe better expose what you want to do before work so we can discuss any issue before you spend time on it). I created the first 4 and you even can tweak WideScreen or change for other one you think is more convenient this days. I think we're on time to change this.

About breakpoints, don't think is so much difficult, I can test a js-define and see if it works as expected if you are interested in it, and if it works ok, then add the rest in that way.

Thanks

PR #816, opened 2020-04-29
- Ensured the latest wagon providers are used by the wagon-maven-plug…

chrisdutz:
…in uses

- Added some properties to fine-tune the way the http wagon provider deploys

REMARK: These properties should probably be disabled or tuned as I switched off paralellism completely ... it might also have an effect on resolving dependencies.


Harbs:
Thanks

PR #822, opened 2020-05-06
Change for custom filter functions receive full object of renderer

raudjcholo:


carlosrovira:
Thanks Raúl for your first royale contribution! :)

PR #827, opened 2020-05-10
Fix - When there is only one button its style should be first and las…

mjesteve:
Fix - When there is only one button its style should be first and last, not only first


carlosrovira:
Many thanks for the fix Maria Jose :)

PR #841, opened 2020-05-25
Add BinaryImage to Jewel

raudjcholo:


carlosrovira:
Hi,
thanks for this contribution. I was reviewing it and looks good, just one thing to improve. It's good to have a Jewel version of BinaryImage so we can apply jewel styles on it, but the composing parts does not change, so we can use Basic pieces. Only in the case we need to add or change something we can extend and create a new class with the changes.
Can you please apply this changes?
Thanks!

Harbs:
What's the use case for adding jewel styles to an image?

If the reason for this is to be able to use the `j` namespace in MXML, there's no need to add classes at all. You can add the following to the Jewel manifest:

`<component id="BinaryImage" class="org.apache.royale.BinaryImage" lookupOnly="true"/>`

carlosrovira:
Hi Harbs,

thanks for pointing that. In Jewel case all components has a CSS rule associated to work consistently cross themes. In this case we still need to add its own and will probably be added to the _image.sass file, so when using runtime theme switching all work as expected.

raudjcholo:
> Hi,
> thanks for this contribution. I was reviewing it and looks good, just one thing to improve. It's good to have a Jewel version of BinaryImage so we can apply jewel styles on it, but the composing parts does not change, so we can use Basic pieces. Only in the case we need to add or change something we can extend and create a new class with the changes.
> Can you please apply this changes?
> Thanks!

Perfect!!

I've already changed it

carlosrovira:
Many thanks! merged! :)

PR #869, opened 2020-06-17
New bead to select an item in the ComboBox of Jewel from a variable

raudjcholo:
This bead allows to select an item by Field. It has two variables: valueField and selectedValue.

- valueField: The String that will be used for comparison field name
- selectedValue: Any kind of object to perform the comparison or select the item with this value in the field.

### DIFF
--- a/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
+++ b/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
@@ -1,0 +1,71 @@
+package com.iest.winplusweb.beads
+{
+  import org.apache.royale.core.IBead;
+  import org.apache.royale.jewel.ComboBox;
+  import org.apache.royale.core.IStrand;
+  import org.apache.royale.collections.CollectionUtils;
+  import org.apache.royale.collections.ArrayList;
+  import org.apache.royale.events.Event;
+
+  public class ComboBoxItemByField implements IBead{
+
+      protected var comboBox:ComboBox;
+
+
+      public function ComboBoxItemByField()
+		{
+		}
+
+        public function set strand(value:IStrand):void
+		{
+			comboBox = value as ComboBox;
+			comboBox.addEventListener("selectionChanged", selectionChangedHandler);
+			updateHost();
+		}
+
+        private var _valueField:String;
+        
+        public function get valueField():String
+        {
+            return _valueField;
+        }
+        public function set valueField(value:String):void
+        {
+            _valueField = value;
+			updateHost();
+        }
+
+        private var _selectedValue:*;
+
+        public function get selectedValue():*{
+            return _selectedValue;
+        }
+
+        public function set selectedValue(value:*):void{
+            _selectedValue = value;
+            updateHost();
+        }
+
+        protected function updateHost():void
+		{
+			if(comboBox && valueField != "" && selectedValue != null){
+			    var aux:* = CollectionUtils.getItemByField(comboBox.dataProvider as ArrayList,valueField,selectedValue);
+                if(aux == null){
+                    comboBox.selectedItem = null;
+                    comboBox.selectedIndex = -1;
+                } else if (aux!==comboBox.selectedItem){
+                    comboBox.selectedItem = aux;
+                }
+            }
+		}
+
+        protected function selectionChangedHandler(event:Event):void{
+            if(valueField != "" && comboBox){
+                var selectedItem:Object = comboBox.selectedItem;
+                if(selectedItem != null && selectedValue !== selectedItem[valueField])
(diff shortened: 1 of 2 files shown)


carlosrovira:
Hi Raúl, thanks for the contribution :)

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
```diff
@@ -0,0 +1,71 @@
+package com.iest.winplusweb.beads
```
yishayw: This looks wrong. Also, I don't see Apache headers in this file.

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
```diff
@@ -0,0 +1,71 @@
...
+      protected var comboBox:ComboBox;
+
+
+      public function ComboBoxItemByField()
+		{
+		}
+
+        public function set strand(value:IStrand):void
+		{
+			comboBox = value as ComboBox;
```
yishayw, line 65: Have you considered casting this to an interface so it can work on strands other than jewel ComboBox? It could be better to place this in Basic so other component sets can use this.

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
```diff
@@ -1,21 +1,65 @@
...
   import org.apache.royale.events.Event;
+  
+  /**
+	 *  The ComboBoxItemByField class is a specialty bead that can be used with
+	 *  any ComboBox control. This bead allows to select an item by field
+	 *  
+	 *  @langversion 3.0
+	 *  @playerversion Flash 10.2
+	 *  @playerversion AIR 2.6
+	 *  @productversion Royale 0.9.4
```
yishayw, line 35: Royale version should be the next released one. Unfortunately, we still don't know what the next version number will be... I suggest using 0.9.8 because that has been the convention so far.

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
```diff
@@ -0,0 +1,71 @@
...
+
+        public function set selectedValue(value:*):void{
+            _selectedValue = value;
+            updateHost();
+        }
+
+        protected function updateHost():void
+		{
+			if(comboBox && valueField != "" && selectedValue != null){
+			    var aux:* = CollectionUtils.getItemByField(comboBox.dataProvider as ArrayList,valueField,selectedValue);
```
yishayw, line 115: CollectionUtils.getItemByField() expects an ICollectionView. It's better to use an interface rather than a conrete class when possible for better reuse.

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
raudjcholo: Ok, perfect!!.

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
raudjcholo: I will change it. Thanks for the feedback

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
raudjcholo: Ok, it's a great idea!!!

Harbs:
Should we revert the merge until the suggestions are implemented?

carlosrovira:
Don't think so. The merge is valid as it's in the current state. What we are asking is for a next iteration to enhance the solution to make it more available for more components.

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controls/combobox/ComboBoxItemByField.as
raudjcholo: Hi:

I have tried to change the code to use an interface but I have not got it to work. 

I will send an email to the list with the code to see if you can help me.

Thanks!

PR #872, opened 2020-06-23
Correction of the text of some constants and 12 new ones.

mjesteve:


Harbs:
The changes don't look right. See here: https://gist.github.com/zaucy/7c735998565b53f1bed9

Does newer versions of Font Awesome use different selectors? If so, maybe it's more appropriate to create new constants for those.

carlosrovira:
Hi  @Harbs, this ones are for FA5 (as the name of the class states. Maybe the ones from zaucy are for FA4?, we have other file for FA4 constants

Harbs:
In that case, doesn't it make sense to rename the constant names as well?

Harbs:
i.e. I'd expect either:
`public static const SORT_ALPHA_ASC:String = "sort-alpha-asc";`
or `public static const SORT_ALPHA_UP:String = "sort-alpha-up";`
If `sort-alpha-up` is the only proper value for FA5, I'd think we should only have the second version.

Harbs:
Is there a reason to mix and match the values that I'm not understanding?

mjesteve:
I thought it would be better to respect the current constants, but really in FA5 there is no "sort-alpha-asc" or "sort-numeric-asc" ... Do I change it?

Harbs:
Yes. I think so. It seems confusing to have a different constant name than what FA requires. 

Thanks for the PR! :)

carlosrovira:
Yeah, const must match strings since otherway are confusingt.
thanks

mjesteve:
> Yeah, const must match strings since otherway are confusingt.
> thanks

¿Is Ok?

Harbs:
Perfect. Thanks!

PR #876, opened 2020-06-26
parameters of type Array throws exception

nihavend:
parameters of type String and XML handled correctly, added support for type Array

### DIFF
--- a/examples/jewel/TourDeJewel/src/main/royale/DateComponentsPlayGround.mxml
+++ b/examples/jewel/TourDeJewel/src/main/royale/DateComponentsPlayGround.mxml
@@ -188,7 +188,7 @@ limitations under the License.
 				</j:CardHeader>
 				<j:CardPrimaryContent>
 					<j:Label multiline="true">
-						<j:html><![CDATA[<p>Following <b>DateField</b> size is not set, so it defaults to 200px.</p>]]></j:html>
+						<j:html><![CDATA[<p>Following <b>DateField</b> size is not set, so it defaults to 162px.</p>]]></j:html>
 					</j:Label>
 					<j:DateField>
 						<j:beads>
--- a/examples/jewel/TourDeJewel/src/main/royale/LayoutsPlayGround.mxml
+++ b/examples/jewel/TourDeJewel/src/main/royale/LayoutsPlayGround.mxml
@@ -198,7 +198,7 @@ limitations under the License.
 
 					<j:Group className="wrapper" width="100%">
 						<j:beads>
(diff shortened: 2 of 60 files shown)


nihavend:
> Are you sure that's what Flex is doing? I thought you said Flex sent something like <> (for an empty XML). 
> Won't this set paramsToSend to null? 

I am sure this is not what Flex is doing, I described below step by step what Flex is doing. Flex setting value to "<>" it is true 
and also it is true that  this will set paramsToSend to null. If it is better to to set it to "<>" than null I can do it.

> I also thought you said that sometimes parameters were passed in. Do you want to handle that case as well? 

If parameters were passed than it is not null and than it is xml or string and handled correctly. Am I missing something ?

> What does Flex output for null and if there are some parameters?

If there are parameters than it can not be null. Am I missing something ?

Here is what flex is doing : 

This is what my app function calls when my internal variable parameter is null : 

 ```
			if(parameter == null) {
				return service.operations[operationName].send();
			}
			
			return service.operations[operationName].send(parameter);
```
Here the function called is  (Operation.as):

`override public function send(... args:Array):AsyncToken`

the value of args : 

image

and this is also the value of "param" which is internal parameter of Operation.

image

The methods returns with `return sendBody(params);` 

sendBody is method of mx.rpc.http.AbstractOperation and the below part of the function is executed : 

```
                    var encoder:SimpleXMLEncoder = new SimpleXMLEncoder(null);                    
                    var xmlDoc:XMLDocument = new XMLDocument();
                    
                    //right now there is a wasted <encoded> wrapper tag
                    //call.appendChild(encoder.encodeValue(parameters));
                    var childNodes:Array = encoder.encodeValue(parameters, new QName(null, "encoded"), new XMLNode(1, "top")).childNodes.concat();                    
                    for (var i:int = 0; i < childNodes.length; ++i)
                        xmlDoc.appendChild(childNodes[i]);

                    paramsToSend = xmlDoc.toString();
```
paramsToSend becomes "<>" here.

image

and the rest of the code creates message object and calls invoke : 

```
        message.contentType = ctype;
        message.method = method.toUpperCase();
        if (ctype == CONTENT_TYPE_XML && message.method == HTTPRequestMessage.GET_METHOD)
            message.method = HTTPRequestMessage.POST_METHOD;
        message.body = paramsToSend;
        message.httpHeaders = getHeaders();
        return invoke(message);
```

aharui:
IMO, Royale should send the same thing as Flex.  Servers may be expecting it.  So I would have Royale set paramsToSend = "<>"

PR #880, opened 2020-07-02
ErrorImage Bead for IImage implement

mjesteve:
The ErrorImage class is a bead that can be used to display an alternate image, in the event that the specified image cannot be loaded.
It will be supported by controls that load the ImageModel bead and implement the IImage interface.

### DIFF
--- a/frameworks/projects/Basic/src/main/royale/org/apache/royale/html/beads/ErrorImage.as
+++ b/frameworks/projects/Basic/src/main/royale/org/apache/royale/html/beads/ErrorImage.as
@@ -1,0 +1,151 @@
+////////////////////////////////////////////////////////////////////////////////
+//
+//  Licensed to the Apache Software Foundation (ASF) under one or more
+//  contributor license agreements.  See the NOTICE file distributed with
+//  this work for additional information regarding copyright ownership.
+//  The ASF licenses this file to You under the Apache License, Version 2.0
+//  (the "License"); you may not use this file except in compliance with
+//  the License.  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+package org.apache.royale.html.beads {
+
+    import org.apache.royale.core.IBead;
+    import org.apache.royale.core.IStrand;
+    import org.apache.royale.events.Event;
+
+    import org.apache.royale.events.EventDispatcher;
+	import org.apache.royale.core.IImage;
+    import org.apache.royale.core.IImageModel;
+  
+  /**
+	 *  The ErrorImage class is a bead that can be used to 
+     *  display an alternate image, in the event that the specified image 
+     *  cannot be loaded.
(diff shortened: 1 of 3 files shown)


frameworks/projects/Basic/src/main/royale/org/apache/royale/html/beads/ErrorImage.as
```diff
@@ -0,0 +1,151 @@
...
+                if(_strand is IImage)
+                {
+                    _hostElement = (_strand as IImage).imageElement as HTMLImageElement;
+                    if(_hostElement){
+                        
+                        _hostElement.addEventListener('error', errorHandler);
+
+                        if(_emptyIsError)
+                        {
+                            (_strand as EventDispatcher).addEventListener("beadsAdded", beadsAddedHandler);
```
yishayw, line 97: Listening to "beadsAdded" is only necessary if this bead is added before its dependencies, in this case the strand's model. Model is usually the first bead to be loaded anyway so this can be simplified by copying the code from beadsAddedHandler here instead.

frameworks/projects/Basic/src/main/royale/org/apache/royale/html/beads/ErrorImage.as
```diff
@@ -0,0 +1,151 @@
...
+        private function beadsAddedHandler(event:Event):void
+        {
+            (_strand as EventDispatcher).removeEventListener("beadsAdded", beadsAddedHandler);
+
+            if(hostModel)
+                hostModel.addEventListener("urlChanged",srcChangedHandler);
+            srcChangedHandler(null);
+        }
+
+        private var _emptyIsError:Boolean = false;
```
yishayw, line 114: pulic attributes should be documented

frameworks/projects/Basic/src/main/royale/org/apache/royale/html/beads/ErrorImage.as
```diff
@@ -0,0 +1,151 @@
...
+            {
+                hostElement.src = _src;
+            }
+        }
+		
+        private function srcChangedHandler(event:Event):void
+        {
+            if(hostModel && !hostModel.url){
+                
+                COMPILE::JS {
```
yishayw, line 135: COMPILE::JS blocks are generally used for platform specific operations. In this case it doesn't look like there's anything js specific in the block.

carlosrovira:
Hi just saw there's another PR #881, should we close this one?

frameworks/projects/Basic/src/main/royale/org/apache/royale/html/beads/ErrorImage.as
yishayw: Good start, but you need to add some annotations to make it visible to IDEs and API Docs. E.g. 

		/**
		 *  Indicates whether the "empty or null" values will be treated as errors and replaced by the indicated src
		 *
		 *  @langversion 3.0
		 *  @playerversion Flash 10.2
		 *  @playerversion AIR 2.6
		 *  @productversion Royale 0.9.8
		 */

mjesteve:
Do I cancel this branch and put all the changes in the PR of 881?
Pending:
- addEventListener 'addbead': As I was saying, the addToParent, in UIBase, loads "first the beads added directly in the mxml tag" and then loads the model. At that point in the code the model is not loaded yet. What am I doing wrong? (I'm going to ask it in the list because I've been looking at it and I don't see any other option)

- COMPILE: JS {}: I try to detect the rest.

Thank you.

yishayw:
I see. In that case listening to the beadsAdded event makes sense. Sorry for the noise.
________________________________

PR #881, opened 2020-07-02
ErrorImageButton Bead for ImageButton controls (Tested in Basic and J…

mjesteve:
The ErrorImage class is a bead that can be used to display an alternate image, in the event that the specified image cannot be loaded.
It will be supported by controls that load the ImageModel bead and expose an imageElement property that references its img element.

The Basic / Jewel ImageButton controls do not share any standard interface for the image, but the Basic ImageButton contains the _imageElement property to save it.
What I propose is to make this public prority and add it to Jewel ImageButton.


mjesteve:
Carlos, this bead is for ImageButton's controls.
I have uploaded the two options separately. I have commented on the @dev list

mjesteve:
Created IImageButton interface (copy of IImage), modified Basic and Jewel ImageButton controls to implement it.

Corrections indicated in # 880:
- I still have doubts regarding the COMPILE :: JS directives. In IImage the get imageElement function is defined for JS only, shouldn't you keep COMPILE :: JS in every function where it appears?
I have left the directive in private function beadsAddedHandler because in all the places where I have seen it implemented it is like this, but I don't know if it will be correct.
I have compiled my fork and TDJ and it doesn't give me errors but I don't know if it will be correct.
- (Uncorrected listener to addBead. Pending response in # 880)

mjesteve:
I'm going to apply the same changes that I applied in the bead that accepted ErrorImage.
If you do not see the interface correct or prefer to do it differently, please tell me.

Thank you.

Harbs:
Why is there  a need to both `ErrorImage` and `ErrorImageButton`? Both beads seem to be identical.

carlosrovira:
Hi @Harbs I saw your commit. I was doing some changes too :), let me know if you has finished so I can do mine.
I'm guessing that ErrorImageButton can be removed too, but I need to do after some safe steps and show Maria Jose some changes to improve. Thanks

Harbs:
I'm done. I was going to work on the release now and I wanted to get this into the current release.

What else did you want to change?

carlosrovira:
I'm simplifying more this bead and adding a default. Going to sync with your changes and merge. Then I'll go over ImageButton since I think too we can rely on ErrorImage

carlosrovira:
@mjesteve I did some changes in ErrorImage to improve code so you can take a look:
- Extends from Bead to avoid IEventDispatcher (that means "removeEventListener" in beadsHandler was removed, but is a better trade off for a Basic bead.
- hotModel stores in _hostModel to avoid making a "getBeadByType"  each time

a part from that:
- Added a default src ("assets/no-image.svg"), so we can avoid to add this property each time.

I'll check later ErrorImageButton, since I think it could be not needed by using just ImageModel that is common in ErrorImage in an extended way, but I  need to check if I'm right.

carlosrovira:
Hi, thanks to @Harbs changes nothing more is needed and `ErrorImageButton` can be removed.
One thing I notice is that the example in TDJ for `js:ImageButton` is not showing the error image, So that needs to be check as a final thing. I have to go out now, if not one do it, I'll go over at later time.

mjesteve:
This afternoon I will study your changes.
I have a single version for Image and for ImageButton Basic / Jewel that detects IImage and IImageButton but I have to look.
Thank you.

PR #887, opened 2020-07-14
ArrayListView: Dispatch lengthChanged binding event if necessary after a refresh

brianraymes:
ArrayListView does not dispatch a "lengthChanged" event after performing a refresh which can alter the length from a filter.

**Steps to Reproduce**

1. _Create a bindable ArrayListView_

```
[Bindable]
public var names:ArrayListView = new ArrayListView(new ArrayList(["A", "B", "C"]));
```

2. _Bind something to the length property_

```
<j:Label text="Length: {names.length}"/>
```

3. _Add a filter and refresh_

```
names.filterFunction = filterFunctionThatRemovesB;
names.refresh();
```

**Expected Results**

Label is updated from `Length: 3` to `Length: 2`.

**Actual Results**

Label is not updated and remains as `Length: 3`.


carlosrovira:
Thanks for you PR Brian! Keep it coming! :)

PR #888, opened 2020-07-21
Crux: Fix a typo where COMPILE::JS should have been used over SWF

brianraymes:
Found a small bug where COMPILE::SWF was used to used to retrieve a value from a map that should have been for JS.


greg-dove:
Thanks @brianraymes - great catch! I suspect that was a copy/paste failure on my part when I added that.

PR #894, opened 2020-08-10
Fixed the example B0015: It was not compiled and the HSlider did not …

mjesteve:
…apply to the List.

A "possible fix" has been implemented: HSliders always modify both containers. Perhaps it would be better if the HSlider modified the TileHorizontalLayout corresponding to the container of the "tab" that is being displayed. I've done something simple ...

Add FontAwesome5IconType constants.

### DIFF
--- a/examples/blog/BE0015_Using_Jewel_TileHorizontalLayout/asconfig.json
+++ b/examples/blog/BE0015_Using_Jewel_TileHorizontalLayout/asconfig.json
@@ -29,6 +29,6 @@
     "additionalOptions": "-remove-circulars -js-output-optimization=skipAsCoercions",
     "files":
     [
-        "src/main/royale/BE0014_Working_with_vector_graphics.mxml"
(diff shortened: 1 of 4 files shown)


carlosrovira:
Hi @mjesteve please don't mix in the same PR different things, since one could be right but the other couldn't, and makes difficult for reviewers to do the review, also, a could happen that a reviewer knows about one of the issues but not the other, so could not end the review and merge it.

You should remove FW icons from this one and add to a different PR.

Thanks!

carlosrovira:
about the BE0015 example,  I must say that I didn't finished it yet, so it could change in the future. Things like config could be good, but I could merge code and when returning in the future to create the blog example could need to change it....or not. Can't ensure until I don't take over it again.

mjesteve:
I don't know how to remove an item from a commit. If that's okay with you, I remove this PR and create 2 new ones. you think?

carlosrovira:
that's ok thanks

PR #895, opened 2020-08-10
jewel-listitemrenderer: Partial revert of data setter so dataChange event can fire when necessary

brianraymes:
This small fix reverts part of a recent change that allows the "dataChange" event to fire when data is changed.


Harbs:
Looks correct to me. @carlosrovira Considering you're more familiar here, I'll leave this to you to merge.

carlosrovira:
Hi @brianraymes and @Harbs, I fixed the problem. While that fixes the issue, it removes the motivation of the change. I detected a double data assignment in renderers due to overrides, so keeping that way, while works, will always generate that problem. The problem is I forgot to add the "dataChange" event. All should be ok now. Let me know if something still fails and thanks for reporting! :). Closing this.

PR #911, opened 2020-09-13
save method from FileReference now operational with a default file name

SolidSoft-Lda:


Harbs:
This commit looks good. Any reason not to merge?

SolidSoft-Lda:
At that time I didn't have write access to the repository so I could only pull request or share code on mailing list. Already merged.

PR #922, opened 2020-10-20
Fixed Jewel ComboBox freeze with SearchFilter

SolidSoft-Lda:
Fixed the following issue:
1. Locate ComboBox on Tour de Jewel components;
2. Go to Local Search Filter sample;
3. Select any item with mouse;
4. Select all text and type "i";
5. The only option will be "Iron Man" and at this point you coun’t select the option or close the list and the application would freeze.

### DIFF
--- a/frameworks/projects/Core/src/main/royale/org/apache/royale/core/Bead.as
+++ b/frameworks/projects/Core/src/main/royale/org/apache/royale/core/Bead.as
@@ -71,5 +71,13 @@ package org.apache.royale.core
       (_strand as IEventDispatcher).addEventListener(eventType, handler, capture);
     }
 
+    /**
+     * Helper function to remove event listener without the need for casting
+     * @royaleignorecoercion org.apache.royale.events.IEventDispatcher
+     */
+    protected function unlistenOnStrand(eventType:String,handler:Function,capture:Boolean=false):void
+    {
+      (_strand as IEventDispatcher).removeEventListener(eventType, handler, capture);
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


carlosrovira:
Hi Hugo,

I could try and that fix the problem in part, checking the code don't understand why it was not working before since seems to be the same. Maybe the remove listener was affecting...

while testing I see that pushing key down is not working too with search filter, that is something we should take a look but I think could be left for other ticket.

About the helper function: I think is important to not mix different things in the same PR. So that should be discussed in other way.

So, since you're a committer, you don't need me to merge, you can do it yourself. I think you can:

- Close this PR
- Merge the Jewel ComboBox fix
- Open a thread in mailing list about the helper function in bead, since @Harbs and others could have something to say. That's because we try to keep core as PAYG as possible so that kind of changes need to be advised.

Thanks!

SolidSoft-Lda:
Hi Carlos,

Yes, in first time the fix was not tested on TDJ. Something that I understand now that should be done for this types of fixs.
However the second time, I did everything (check out Royale source code, compile, etc ...) and tested so, yeah, something was in the middle.

The helper function on this PR was something that I did wrong. However there is another PR just for that.

About commit directly on apache royale:
Is have an e-mail (gmail) that I use to mailling list group, another e-mail (solidsoft) for my git account and there is also the new apache e-mail that was created (I don't have the account setup and password yet).
I should use the apache e-mail as user on github for apache royale or can I use my soldisoft account ?
If should be with the new apache e-mail, where can I recover the password ?

carlosrovira:
Hi Hugo, I'll contact you in private to see how to commit

Issue #641, opened 2019-12-29
Fix PAYG violations and code debt


Harbs:
We have built up code debt by not being strict enough in PAYG. This issue is an attempt to track PAYG and code inflation problems and come up with solutions to fix them.

I'm using HelloWorld to analyze and improve PAYG issues for now.

Identified issues:
- [x] Dependency on OSUtils
- [x] Dependency on StringUtil
- [ ] Minified code has many duplicate prototype definitions. You can find them by using the following RegExp: `(\w{0,})(\.prototype\.)(\w{0,})=\1\2\3;` Removing these reduced the compiled code of HelloWorld from 97,715 bytes to 96,539 bytes.
- [ ] Dependency on goog.events adds code dependencies on a lot of unnecessary code like browser sniffing. Much of this is even for older browsers that Royale does not support.
- [ ] Reflection data is included for all classes used -- even if that reflection data is never actually used in the app. We should look for ways to eliminate reflection data not actually needed.
- [ ] All package paths and classes are exported. This causes an non-trivial amount of code to be included.
- [ ] All public functions are exported. This prevents dead code removal for functions not used and prevents function renaming. We should examine whether this can be improved.

Harbs:
I fixed the first two problems today. That resulted in an improvement of a few KB.

Harbs:
I'm not sure why we have the duplicate function definitions. I'm pretty sure this was introduced at some point and it seems to happen at the minification stage.

Harbs:
I don't remember why public functions are exported. getters need to be exported for MXML support, but I'm not sure why that would apply to functions.

greg-dove:
Hi Harbs,

A couple of quick comments:


Reflection data is included in debug build as it should be, but last time I
checked I think it was completely omitted in HelloWorld release build.
Please let me know if that is not the case.


If you have a static utils class, you can eliminate reflection and export
selectively on that class by annotating the class itself with
@royalesuppressexport. I did this already with js Language, for example.
This approach should not be used without consideration about how it limits
use of the class. But I think it would probably be low risk for sets of
utility functions. It would not be advisable for the general case with a
class that has public static costs for example... they may need to be
accessed with string keys in release build (if used in mxml bindings for
example) and therefore continue to need 'export'

Harbs:
Here's is some minified code which I believe comes from CSSUtils:
````
[...]

PR #923, opened 2020-10-20
Helper function to remove event listener without the need for casting

SolidSoft-Lda:


Harbs:
Because of PAYG, we don't want to add a remove function directly to Bead.

A stand alone utility function is better for this.

Thanks!
Harbs

carlosrovira:
Hugo, is easy to change to a stand alone function, you have lots of examples in Royale.
Thanks

SolidSoft-Lda:
Yeah, it's not a big issue for me. I had the feeling that was incomplete because we already have the add event listener helper anyway, but, it's ok for me. I understand that PAYG have rules.

Harbs:
I'm hoping we can get to the point where PAYG wil be helped by the compiler and unused methods can be stripped out of compiled apps. We're not yet sure that's going to be possible, so we need to be stricter than would otherwise be ideal.

Here's more info for your reading pleasure. 😄 https://github.com/apache/royale-asjs/issues/641

SolidSoft-Lda:
Very cool :)

carlosrovira:
yeah that would be super awesome to get implemented in the compiler :)
I read that some js tech out there already does it, so I think it's completely possible for Royale too.
Just we need the proper muscle to get it done as always :)

PR #942, opened 2020-11-21
mx.messaging.channels.URLVariables: implemented for HTTPService.request in MXRoyale / SparkRoyale

estanglerbm:
HTTPService is unable to send HTTP GET parameters, specified in the request member, without an implementation of URLVariables.  This trivial implementation leverages the existing org.apache.royale.net.URLVariables.


carlosrovira:
This change seems ok to me but don't have a test code to check. Hope others that have more easy than me some way to do a quick check. Someone has some opposition to merge it?

Harbs:
It looks like it's relying on the base class for everything. Why any overrides at all?

estanglerbm:
Since it's the lowest level of the mx hierarchy, I thought it would be good to explicitly state the API.  But sure, they could be removed.

greg-dove:
Thanks very much for your contribution. I added the org.apache.royale.net.URLVariables implementation at some point and did not look in MXRoyale at the time. The URLVariables implementation is useful in general outside of the flex emulation code. So the main issue here is that the MXRoyale 'implementation' is not really an implementation, but an empty 'stub' class. Here's what I'd like to suggest: instead of adding the subclass in MXRoyale, we remove it and substitute the working version from the Network library. It is only used in one place inside DirectHTTPChannel and can be considered an 'internal' utility support only. This would get that part working like it should, and removes a redundant layer of inheritance. And if you need the class in any local code it is still available for import from the Network library. Would you be happy with that?

greg-dove:
As a quick follow-on to my last comment:
If you are happy with that suggestion, there are two options - you could create a new pull request for that change, that does the following:
```
1.
delete the URLVariables class in MXRoyale
2.
remove this from the top level MXRoyaleClasses.as:
import mx.messaging.channels.URLVariables; URLVariables;

3.
in mx.messaging.channels.DirectHTTPChannel
add this inside the COMPILE::JS import block:

import org.apache.royale.net.URLVariables;
```

If you want to do that, I will happily accept the request this gets you (rightfully, given that the solution derived from your efforts) into the project's commit history, so your continued contributions can hopefully support you gaining committer rights in the future - we're definitely keen to welcome more active community members like yourself.

The other option is that if you prefer otherwise, just let me know and I will make this change in the next 24 hours.

estanglerbm:
Replaced with pull request #948

PR #945, opened 2020-11-22
XMLList: added localName()

estanglerbm:
Existing Flex code can call localName() on an XMLList object.

Not sure if there is some automatic conversion that needs to be implemented, or if you prefer that people call toXML().localName(), but adding localName() to XMLList seems like the easiest way to get existing code to compile unchanged.

### DIFF
--- a/frameworks/projects/XML/src/main/royale/XMLList.as
+++ b/frameworks/projects/XML/src/main/royale/XMLList.as
@@ -1454,6 +1454,15 @@ package
 			throwError("Incompatible assignment of XMLList to XML");
 			return null;
 		}
+
+		public function localName():String
+		{
+			if (isSingle())
+				return _xmlArray[0].localName();
+			
+			throwError("Incompatible assignment of XMLList to XML");
+			return null;
+		}
 	}
 }
 


Harbs:
Thanks for this. I must have missed that method. Does this throw an error in Flash if the XMLList is empty?

The rest of these methods do not throw errors so I'm trying to figure out the best behavior.

Also, a unit test for this would be good.

@greg-dove Do we already have unit tests for single item XMLList access of XML methods?

greg-dove:
@Harbs I can't recall, I can check first thing tomorrow local time.

To answer your other question:

TypeError: Error #1086: The localName method only works on lists containing one item.
	at XMLList/http://adobe.com/AS3/2006/builtin::localName()
	at Untitled_2_fla::MainTimeline/frame1()

Harbs:
I'm guessing the rest of the methods throw as well (and we should add that). It's probably better to use `org.apache.royale.debugging.assert` so the code to throw the error is stripped out in release.

Thoughts?

Harbs:
Merging this to get this change in the upcoming release. We can deal with error handling in the future.

PR #948, opened 2020-11-22
Removed mx.messaging.channels.URLVariables, and switched to using org.apache.royale.net.URLVariables for HTTPService.request in MXRoyale / SparkRoyale.

estanglerbm:
HTTPService is unable to send HTTP GET parameters, specified in the request member, without an implementation of URLVariables.

After doing a trivial implementation that leveraged the existing org.apache.royale.net.URLVariables, @greg-dove suggested that I remove the MX version of URLVariables and just use the Royale version.

This pull request replaces #942.


estanglerbm:
@greg-dove:  Thanks for the comments on the original pull request.  It's fine either way.  As long as I don't have to custom build MXRoyale for this change in future releases!

greg-dove:
Thanks heaps. 

>  As long as I don't have to custom build MXRoyale for this change in future releases!

We definitely prefer working code, so your contribution is most welcome!

PR #957, opened 2020-11-26
spark.components.Application:  Implemented parameters for Spark Application (copied from MX Application)

### LINKED ISSUE #956: ReferenceError: Error #1056: Cannot create property parameters on ... in Flash Debug Player for Spark Application
Spark Application, unlike MX Application, lacks the parameters member, so Flash Debug Player throws:

`ReferenceError: Error #1056: Cannot create property parameters on ...`

estanglerbm:
Fixes #956


Harbs:
Thanks!
It might be an idea to change MX and Spark Applications to use beads for this in the future. Not sure if that complexity would add anything though.

Issue #953, opened 2020-11-25
VerifyError: Error #1053 in Flash Debugger because Royale compiler allows incompatible function signatures


estanglerbm:
flashplayer_32_sa_debug.exe throws this error:

`VerifyError: Error #1053: Illegal override of A in a.A.`

With this:

```
public interface B
{
	function foo(c:Object):void;
}

public class A implements B
{
	public function foo(c:String):void {}
}
```

Royale mxmlc happily compiles this.  Flex mxmlc will give this compile error:

`Error: Interface method foo in namespace a:B is implemented with an incompatible signature in class a:A.`

Why is this important?

If you have:

```
import org.apache.royale.events.Event;

class ... implements IEventDispatcher
{
...
	public function dispatchEvent( event : Event ) : Boolean { ... }
}
```

Royale mxmlc compiles it just fine.  And it runs fine in JS.  But in Flash Debugger, it throws VerifyError: Error #1053.

The correct definition is:

```
COMPILE::SWF
{
    import flash.events.Event;
}

class ... implements IEventDispatcher
{
...
	COMPILE::SWF
	public function dispatchEvent( event : flash.events.Event ) : Boolean { ... }

	COMPILE::JS
	public function dispatchEvent( event : Object ) : Boolean { ... }
}
```

And this kind of problem is hard to track down.

joshtynjala:
I can reproduce the VerifyError. That's not good.

Using the `-allow-subclass-overrides=false` compiler option when compiling your app may help here.

Unfortunately, making `-allow-subclass-overrides=true` the default seems to have been intentional, and the Royale framework now takes advantage of it widely. It may be very difficult or impossible to fully unravel at this point.

PR #960, opened 2020-11-29
Fix VerifyError: Error #1053 in Flash Debug Player due to mismatched types

estanglerbm:
Fix VerifyError: Error #1053 in Flash Debug Player due to wrong IDataInput / IDataOutput used in mx.messaging.messages.* classes, and due to mismatch of IFactory type between spark.components.DataGroup and IItemRendererProvider.  (Both are the problem described in issue #953.)

For mx.messaging.messages.*, the conditional imports for IDataInput / IDataOutput (for IExternalizable) used elsewhere were not used here.  The compiler didn't complain.

For DataGroup, there is a problem that it defines set/get itemRenderer with mx.core.IFactory, but it's implementing IItemRendererProvider that uses org.apache.royale.core.IFactory.  This is hard to reconcile, so I used SWFOverride.  (FYI, mx.controls.listClasses.ListBase uses org.apache.royale.core.IFactory when implementing IItemRendererProvider, so there's some inconsistency.)

### DIFF
--- a/frameworks/projects/MXRoyale/src/main/royale/mx/messaging/messages/AbstractMessage.as
+++ b/frameworks/projects/MXRoyale/src/main/royale/mx/messaging/messages/AbstractMessage.as
@@ -21,8 +21,16 @@ package mx.messaging.messages
 {
 
 import org.apache.royale.utils.BinaryData;
-import org.apache.royale.utils.net.IDataInput;
-import org.apache.royale.utils.net.IDataOutput;
+
+COMPILE::JS {
+	import org.apache.royale.utils.net.IDataInput;
+	import org.apache.royale.utils.net.IDataOutput;
+}
+COMPILE::SWF{
+	import flash.utils.IDataInput;
+	import flash.utils.IDataOutput;
+}
+
 import org.apache.royale.reflection.getQualifiedClassName;
 
 import mx.core.mx_internal;
--- a/frameworks/projects/MXRoyale/src/main/royale/mx/messaging/messages/AcknowledgeMessage.as
+++ b/frameworks/projects/MXRoyale/src/main/royale/mx/messaging/messages/AcknowledgeMessage.as
@@ -20,8 +20,14 @@
 package mx.messaging.messages
 {
 
-import org.apache.royale.utils.net.IDataInput;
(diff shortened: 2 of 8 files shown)


carlosrovira:
Hi @greg-dove could you take a look to this PR?
Thanks in advance for your help on this

Harbs:
I don't see any harm in this change, but I'm concerned that it's needed.

`org.apache.royale.utils.net.IDataInput` extends `flash.utils.IDataInput` in the `SWF` compile, so I would not think you'd get mismatched types. If there are mismatched types, it feels like it should be fixed in `org.apache.royale.utils.net`.

estanglerbm:
Everywhere else that IDataInput / IDataOutput is used, there are conditional imports.  It was just missed in this area, as far as I can tell.  So this isn't something new.

Flex doesn't allow function overloading like you describe;  it throws `Error: Incompatible override.`

That is, it doesn't allow:

```
package a
{
        public class Y {}
        public class X extends Y {}

        public class B { public function foo(c:Y):void {} }
        public class A extends B { override public function foo(c:X):void {} }
}
```

The IDataInput / IDataOutput changes are just copying over the conditional imports from other areas.

The DataGroup change is something new.

greg-dove:
> Hi @greg-dove could you take a look to this PR?
> Thanks in advance for your help on this

Sorry all, I can't look into this more deeply until this coming weekend earliest. But the Fixes for IDataInput/IDataOutput seem correct to me. 
This situation with swf was probably overlooked because I was verifying Royale JS against Flex builds and not against swf builds of the same thing in Royale for actual remoting tests. (For all the low level tests with AMFBinaryData, that is done with both swf and JS in Royale.)
The need for dual imports is valid here, because there are interfaces with method signatures that have arguments that reference other native interfaces in their types in swf, so you can't really 'override' them. Even if it was possible, the effort to make all this 'one' single api to avoid the conditional imports is IMO not worth it. Probably the swf stubs that I added thinking I could do that should be removed, so it is not confusing, but I'm just musing without spending time checking to see.... I can look into it more on of after this next weekend, or if others want to look at that earlier, please go for it!

Harbs:
I merged it for now. @greg-dove Please see if there's another way to go about this when you have a chance.

PR #969, opened 2020-12-01
Implemented JS side of XMLNotifier, ObjectMap.forEach, and many XML notifications.

estanglerbm:
1.  Needed to enumerate an ObjectMap, but WeakMap doesn't have enumeration.  So added a constructor option to force Map to be used instead of WeakMap if enumeration is needed (thought it would be better than subclassing just for that option), and implemented forEach().  Choose forEach() due to the wider browser support, over iteration.  Didn't implement forEach() for the really old browsers.

2.  Implemented the JS side of XMLNotifier.  XMLNotifier multiplexes XML notifications.  A complication is that JS doesn't have arguments.callee due to strict, so added an extra first arg to the internal notification function that communicates between XML and XMLNotifier, only on JS (for SWF, still uses the original method).

3.  Implemented many (but not all) XML notifications for JS.  They are sent for the current node only if setNotification() was called.  For the bubbling-up of notifications through the parent, need to set static XML.recursiveNotify = true (it is disabled by default).  Tested JS notification behavior against Flex, for a few levels of nodes and attributes.  Have not tested complex QNames.

### DIFF
--- a/frameworks/projects/Core/src/main/royale/org/apache/royale/utils/ObjectMap.as
+++ b/frameworks/projects/Core/src/main/royale/org/apache/royale/utils/ObjectMap.as
@@ -42,9 +42,10 @@ package org.apache.royale.utils
 
     public class ObjectMap
     {
-        public function ObjectMap(weak:Boolean=false)
+        public function ObjectMap(weak:Boolean=false,forceEnumerable:Boolean=false)
         {
             _weak = weak;
+            _forceEnumerable = forceEnumerable;
             COMPILE::SWF
             {
                 _map = new Dictionary(weak);
@@ -58,7 +59,7 @@ package org.apache.royale.utils
         COMPILE::JS
         private function makeMap():void
         {
-            if(_weak && typeof WeakMap == "function")
+            if(_weak && !_forceEnumerable && typeof WeakMap == "function")
             {
                 _map = new WeakMap();
                 assignFunctions();
@@ -78,6 +79,7 @@ package org.apache.royale.utils
         }
 
         private var _weak:Boolean;
+        private var _forceEnumerable:Boolean;
         private var _map:Object;
         private var _usesObjects:Boolean = false;
 
@@ -93,6 +95,23 @@ package org.apache.royale.utils
         public function delete(key:Object):void
         {
             delete _map[key];
+        }
+
+        /**
+         *  Perform function for each pair.
+         *  @langversion 3.0
+         *  @playerversion Flash 10.2
+         *  @playerversion AIR 2.6
+         *  @productversion Royale 0.9.8
+         * 
+         */
+        COMPILE::SWF
+        public function forEach(callback:Function,thisArg:Object=null):void
+        {
+            for (var key:Object in _map)
+            {
+                callback(_map[key],key,_map);
+            }
         }
 
         /**
@@ -173,6 +192,10 @@ package org.apache.royale.utils
              *  @royalesuppresspublicvarwarning
              */
             public var delete:Function = objectDelete;
+            /**
+             *  @royalesuppresspublicvarwarning
+             */
+            public var forEach:Function = objectForEach;
         }
 
         COMPILE::JS
@@ -182,6 +205,7 @@ package org.apache.royale.utils
             this.has = _map.has.bind(_map);
             this.set = _map.set.bind(_map);
             this.delete = _map.delete.bind(_map);
+            this.forEach = (_weak ? null : _map.forEach.bind(_map));
         }
 
         COMPILE::JS
@@ -228,6 +252,12 @@ package org.apache.royale.utils
             }
         }
         COMPILE::JS
+        private function objectForEach(callback:Function,thisArg:Object=null):void
+        {
+		// if (_weak) throw ...
+		trace("ObjectMap::objectForEach() not implemented in this browser");
+        }
+        COMPILE::JS
         public function clear():void
         {
             if(_usesObjects)
@@ -243,6 +273,5 @@ package org.apache.royale.utils
                 assignFunctions();
             }
         }
-
     }    
 }
--- a/frameworks/projects/MXRoyale/src/main/royale/mx/utils/XMLNotifier.as
+++ b/frameworks/projects/MXRoyale/src/main/royale/mx/utils/XMLNotifier.as
@@ -94,7 +94,7 @@ public class XMLNotifier
                                                      value:Object,
                                                      detail:Object):void
         {
-            var xmlWatchers:ObjectMap = (callee as Object).watched;
+            var xmlWatchers:ObjectMap = callee["watched"];
             if (xmlWatchers != null)
             {
                 xmlWatchers.forEach( function(truevalue:Object,notifiable:Object,map:Object):void {
@@ -116,7 +116,7 @@ public class XMLNotifier
                                                      detail:Object):void
         {
             var callee:Function = notificationFunction;
-            var xmlWatchers:ObjectMap = (callee as Object).watched;
+            var xmlWatchers:ObjectMap = callee["watched"];
             if (xmlWatchers != null)
             {
                 xmlWatchers.forEach( function(truevalue:Object,notifiable:Object,map:Object):void {
--- a/frameworks/projects/MXRoyale/src/main/royale/mx/utils/XMLNotifier.as
+++ b/frameworks/projects/MXRoyale/src/main/royale/mx/utils/XMLNotifier.as
@@ -20,10 +20,7 @@
 package mx.utils
 {
 
-COMPILE::SWF
-{
-import flash.utils.Dictionary;
-}
+import org.apache.royale.utils.ObjectMap;
 import mx.core.mx_internal;
 import mx.utils.IXMLNotifiable;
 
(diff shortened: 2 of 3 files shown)


estanglerbm:
Updated after some more testing with XMLListAdapter.

greg-dove:
I have some WIP on an alternate implementation of this. I had started, but did not finish. The approach I took was to subclass XML (in as3 this is strictly not possible, but we can in js). If the notification support was requested, it swapped the prototype (again very js-ish) of the current XML instance to use overridden methods that support the notifications - if setNotification is subsequently set to null it swaps back to the lighter base class instance. I did not think (or know) about these bubbling. Your approach sounds comprehensive.
I've encountered the same issues with non-iterable WeakMap/ Map (forEach needed by IE11) etc, so I know the decision path you went through there!  I will come back and attach my WIP here just in case here you want to take a look at it - I'd be happy to look through your code, but I can't spend time on this until this coming weekend I'm afraid. Hopefully someone else can. I also wrote a bunch of unit tests for what I was working on - I will drop that here also - either later today or before I start work tomorrow morning local time

Harbs:
Nice. @greg-dove I'm assuming you're going to merge this PR considering this is an area you worked on.

estanglerbm:
Changes to XML notification after I did more data type analysis, more tests with XMLList, and more tests with setChildren().

Harbs:
I just noticed that this has changes to XML. How (if at all) does this effect XML performance and memory consumption.

I was very sensitive to performance when implementing XML.

estanglerbm:
There are two new members in XML:

```
		private var _notification:Function;
		private var _internalSuppressNotify:Boolean;
```

Most of the changes are sprinkling of calls to xml$_notify().  That new function only calls the notification function if it's been set (through a previous call to setNotification()). It will bubble up to the parent only if the global variable recursiveNotify is set to true (default is false).  So if you never call setNotification() and you never set recursiveNotify, there is just a call to a do-nothing function in each of the various operations.

The changes in setChildren() are a little bit more, only because of the specific legacy behavior of Flex.  But not a lot more, in my opinion.

If you're using XMLNotifier, then each XML will get its own notification function allocated by XMLNotifier (that's how it works).  And each notification function will hold / release a reference to the watcher(s), through watchXML() and unwatchXML().  So at any point in time, there could be a lot of XMLs with notification functions and references to one or more watchers.  But all that stuff with XMLNotifier is part of the price of using that, if you want to use it.

I have not run benchmarks of before vs. after on XML in JS, beyond using it in a real app with a lot of XML.  I did try to have little impact while still making changes that wouldn't be fragile (hence the xml$_notify() function instead of inline).

greg-dove:
@Harbs, @estanglerbm  I will take a look at this over the weekend.

estanglerbm:
Fixed runtime problem in XMLNotifier in release mode.

Harbs:
@greg-dove Have you had a chance to review this?

greg-dove:
> @greg-dove Have you had a chance to review this?

@Harbs I have not. I am caught in a cycle of extreme work hours that has been going for much longer than is healthy. I will force a break in that this coming weekend and look then. I do have an alternative approach, but I had not factored in 'bubbling' with the approach I started. However I also have the beginnings of a set of unit tests that I can check both approaches against.

My apologies for the delay, @estanglerbm, this is quite an unusual circumstance for me.

estanglerbm:
No worries.

greg-dove:
Thanks for your patience. I took a quick look just now.
There are 3 unit tests that no longer pass in JS after the changes. Probably you did not figure out how to run those yet. 
For me to run those, I just need to ensure that PLAYERGLOBAL_HOME is defined (as a parent folder with versions subfolders containing playerglobal.swc ... e.g. %PLAYERGLOBAL_HOME$/11.1/playerglobal.swc ). Then cd to frameworks/projects/XML and run ant in that folder with all other normal royale env vars set. It should build both swf and js swcs for XML, and then run both swf and js tests (the same tests run in js and swf, but obviously this is mainly to use the swf version as a reference for language level stuff). Without PLAYERGLOBAL_HOME set I don't think any tests are run, but I have not checked to see why.
I have added a new start of tests for XML notification - this is now in develop, but not compiled in (because it obviously will fail in current develop). These tests are just for the language level implementation in XML notification, not the mx/emulation stuff. I did not add anything for bubbling tests yet, and I was only exploring this to figure out ways to do it in the past. And the tests are only a start I expect. I did push where I had gotten to as WIP to a branch (https://github.com/apache/royale-asjs/tree/XML_Notification_prototype_swap) which you should feel free to harvest if you feel you can use anything from it. The general principle I was aiming for there wass to make the runtime implementation of XML notification available for dead-code elimination if it was never used, in case you are wondering about the 'strangeness' of it.

At the moment I don't see the same initial notification tests passing which do pass on the Flash player, so I am not sure to do about that. Maybe the goal of your PR is more focused on the emulation support and not at the language (XML) level, which is probably sufficient, because I suspect it is how most people would use this feature.
I think that if the stringification tests that currently have errors in your branch could be fixed then I I think it is probably good for merge, so long as others are fine with that. Thanks again for your efforts on this. I know how time-consuming testing this stuff can be....

estanglerbm:
JS (and SWF) tests are passing for me (though, I have to build manually before running the tests normally):

```
[royaleunit] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.369 sec
[royaleunit] Suite: flexUnitTests.xml.XMLTesterGeneralTest
[royaleunit] Tests run: 31, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.813 sec
[royaleunit] Suite: flexUnitTests.xml.XMLLiteralTest
[royaleunit] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.369 sec
[royaleunit] Suite: flexUnitTests.xml.XMLNamespaceClassTest
[royaleunit] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.246 sec
[royaleunit] Suite: flexUnitTests.xml.XMLTesterStringifyTest
[royaleunit] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.861 sec
[royaleunit] Suite: flexUnitTests.xml.XMLTesterNamespaceQueries
[royaleunit] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.369 sec
[royaleunit] Suite: flexUnitTests.xml.XMLNamespaceTest
[royaleunit] Tests run: 12, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.476 sec
[royaleunit] Suite: flexUnitTests.xml.XMLListTesterGeneralTest
[royaleunit] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.230 sec
[royaleunit] Suite: flexUnitTests.xml.XMLQNameTest
[royaleunit] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.738 sec
[royaleunit]
[royaleunit] Results :
[royaleunit]
[royaleunit] Tests run: 77, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.471 sec
[royaleunit]
...
     [echo]
     [echo] XMLJS.test RoyaleUnit Release build tests: PASSED
     [echo] XMLJS.test RoyaleUnit Debug build tests: Not tested
```

Are you using the latest changes on the branch?  After the initial PR, I checked in some fixes.

estanglerbm:
And for your new set of tests, I made some comments on your commit on some adjustments that may need to be made, including:

- Need for xx.toXMLString() on the targets, and handling various types for value/detail.
- How my XML.setNotification() callback requires an extra first parameter (callee:Function) on JS only (and why).

estanglerbm:
After building the tests a different way, there may be 3 existing tests (1 in stringify, 2 in general) that are failing because _name becomes null (due to existing code) in XML.setName(), and the new code doesn't anticipate that (it calls _name.toString() in one of the callback args).  I need to look into if that's a real error;  if so, it's a quick fix.

greg-dove:
Thanks, yes this is what I saw. One quick question... I did not look at the mx emulation support, and I take your point about this being undocumented. But would it be possible to have the method signature put the 'callee' argument last, to keep the first 5 the same? Just on the off-chance that someone is using this in custom code somewhere.... . Thanks for the info about the tests. I only started this stuff about 6 months back and stalled, so if you're game to fix them and switch them on for your implementation, please don't hold back! Sounds like this can be merged pretty soon.

greg-dove:
btw, just a quick comment about the tests. You should have the junit style output available somewhere after they are run, with information on failures. But if you see the failed js test in the browser, you can append : '?local' and refresh it. You should then have some helpful info in the browser console, including stack traces which help track down issues. This is of course much easier with the debug build compared to the release build...

estanglerbm:
Fixed the existing tests (i.e. stringify), which were failing because setName() is called within parseXMLStr() for processing instructions, and that was unexpected (in that case, existing _name == null).

Still looking at other things...

estanglerbm:
Moved the callee argument to the XML notification callback to be the last parameter, as @greg-dove asked, and simplified XMLNotifier.initializeXMLForNotification().

But existing code, if it's out there, still needs to add the extra parameter to their callback function (and it can be default null).  [This change allows you to use the same function for both JS and SWF, with minimum changes.]

estanglerbm:
Fixed 2 of the 6 failing new tests (XMLNotificationTest).

The remaining 4 failures, I believe, are due to items that I documented in XML.setNotification() as not being done yet (for the 4 failing tests:  textSet, nodeChanged, namespaceSet, textSet).

estanglerbm:
Test-related:
- Implemented XML notifications textSet, namespaceAdded, and namespaceSet
- Added addNamespaceInternal() to help implement namespaceAdded notification in a reentrant way

Non-test-related:
- Fixed a type returned in nameSet notification
- Fixed order of nodeRemoved notifications in some cases of setChild(), to match Flex behavior.  It turns out that Flex lies in the timing of these notifications;  it will send out nodeRemoved notifications _before_ setChild() has actually removed the nodes, at least in some cases.

Now, only one test is failing in XMLNotificationTest, and that's related to nodeChanged, which is not implemented yet.  (namespaceRemoved is also not implemented yet.)

estanglerbm:
Implemented XML notification namespaceRemoved.

That just leaves nodeChanged to be implemented.

estanglerbm:
Fixed behavior of nodeAdded and nodeRemoved when working with XMLList arguments, to match Flex behaviors, and eliminated extra loops for XMLList notification behaviors by default.

There are some XMLList notifications in Flex that involve nodeChanged that are still not handled.

Harbs:
Where are we holding with this PR? I'd like to restart the release process next week and it would be good to have this included.

Harbs:
@greg-dove

greg-dove:
@Harbs @estanglerbm sorry for delays. I will definitely get to it this weekend.

greg-dove:
Thanks again for all your efforts here. Sorry for the delay.

PR #970, opened 2020-12-01
Implemented MX Tree itemOpen event and isBranch tree item attribute.

estanglerbm:
isBranch was a peculiar built-in thing in Flex.

Also fixed opening nodes with no children (which is allowed in Flex).


estanglerbm:
Commented out some traces.

estanglerbm:
Fixed wrong event name.

PR #975, opened 2020-12-02
Fix MX Tree display problems with dynamic dataProvider, and implement Tree.isItemOpen()

estanglerbm:
Fixed display problems in MX Tree when dataProvider is XMLListCollection and you do:

```
mydp.source = mynewxmllist;
mydp.refresh();
```

Also implemented Tree.isItemOpen().


estanglerbm:
Added support for dynamic updates to MX Tree via XML notifications.  (The little dance of XML notifications between XML, XMLListAdapter, and HierarchicalCollectionView works now for JS.)

Fixed problem in MX HierarchicalCollectionView.calculateLength() due to lack of support for indexed array syntax ([]) for ICollectionView.

PR #1004, opened 2020-12-13
Several suggestions for js\bin\mxmlc.bat

estanglerbm:
o Added setlocal to mxmlc.bat files to prevent ROYALE_HOME, etc., from being set in the caller environment.

o Increased max memory for java to 1024 MB to prevent out-of-memory during release JS builds on modest app, and increased minimum memory to 128 MB to speed up compile by 5-10% on modest machine.

o **Adjusted ROYALE_COMPILER_HOME to be consistent with build.xml files (...\js).**  Note that the command-line define for "royalecompiler" seems to be ineffectual (so the change here should not be a problem for that);  there is no System.getProperty("royalecompiler") or such in the .java compiler code.


estanglerbm:
Just for completeness, the out-of-memory error (with 512 MB max) is:

``Exception in thread "main" java.lang.OutOfMemoryError: GC overhead limit exceeded``

1 GB max doesn't seem like an unreasonable amount these days.

PR #1023, opened 2020-12-30
Fix 2 AMF issues in JS build:

SemiConscious:
This fix attempts to fix two problems I found when when making calls to our AMF service:

1) pass in the AMF URI, as currently it's always NULL, and set it in the messageBody
2) fix the 'params' parameter in AMFNetConnection.call, as otherwise it's not possible to send in an array of optional parameters

My main concern in committing this back to the repo is that it changes the signature of `AMFNetConnection.call()`. This is only one solution to the problem. It may be possible to fix the issue by using `.apply()` on the call to that function from `NetConnection.call()` - I will investigate it but I think it would still require some contortions in `AMFNetConnection.call()` to get the desired effect.

The other thing to consider is that `NetConnection.call()` has a different signature in royale compared to flex. I assume this was done for a reason so I didn't change that ... but it matches `AMFNetConnection.call()`. If not then reverting to the flex signature would make existing AMF services easier to port.

Anyway, by all means have a look and let me know your thoughts.

### DIFF
--- a/frameworks/projects/MXRoyale/src/main/royale/mx/net/NetConnection.as
+++ b/frameworks/projects/MXRoyale/src/main/royale/mx/net/NetConnection.as
@@ -150,7 +150,8 @@ public class NetConnection extends EventDispatcher
         var ncResponder:org.apache.royale.net.Responder = 
                         new org.apache.royale.net.Responder(responder.resultFunction,
                                                               responder.faultFunction);
-        nc.call(null, ncResponder, message);
+        // Jim P: pass the AMF target URI in here 
+        nc.call(thisObject as String, ncResponder, message);
     }
 }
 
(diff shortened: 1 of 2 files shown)


carlosrovira:
Hi, many thanks for the PR. It seems ok to me, but we have a 24h delay so others can see if there's something about it. @greg-dove can you take a look? 

Please for further PRs don't mix different issues, so reviewers can check one as ok and other not, and we'll be not able to merge since both are mixed in a single one PR. So always separate each issue in its own PR.

carlosrovira:
Hi @SemiConscious, a user on mailing list (Safrizal) is complaining about AMF broken. I can only think on this PR could be the cause of the problem. I think the best will be to revert this one and discuss on list about the problem. Can you see this problem with Safrizal to reach for the right solution? Thanks

carlosrovira:
@SemiConscious the thread in mailing list is "latest build, amf problem..."

PR #1024, opened 2020-12-31
Fix DataBindingBase.deferredBindings supporting only a single (the last) property

stanislaw89:
Currently, `DataBindingBase.deferredBindings` supports only a single binding per property, i.e. in a case like
```
<j:Button text="{...}" visible="{...}"/>
```
Binding for "text" is added but immediately gets overridden by "visible" binding. As a result, only the last binding works.


carlosrovira:
Thanks for this patch!. seems ok to me. I'll be merging if others not oppose in the 24h frame we have

PR #1025, opened 2020-12-31
Speed up XML parsing of larger datasets by 30-50%

estanglerbm:
Speed up XML parsing of larger datasets by 30-50% by using faster DOM APIs, shorter string lookups, eliminating unnecessary string copies, and other changes.

The original elapsed time was about 80% in post-text-parsing code (node traversal, building XML objects, etc.) and 20% in text-parsing code (parseFromString).  Now, the time in text-parsing code dominates.

The hoops to use Node.getAttributeNames()--in addition to requiring a newer browser--is partly due to this version of GCC not supporting it (was added around 2017, I think).

One (not large) bottleneck is the node traversal, which may be visiting each node twice (due to recursive + nextSibling;  I just wrote a faster version of the existing traversal).  There may be some speedup by using TreeWalker to do nextNode() traversal and using a map to hook up the parent relationships.  But not worth the effort, at the moment.

Another small bottleneck is the trim for whitespace nodes.  A filter during TreeWalker or during parsing would help here, to just eliminate the whitespace nodes from the traversal.

The big bottleneck (by far) is now DOMParser.parseFromString().  I had mixed results using createContextualFragment(), and that executes script, so not ideal.  I think it may be best to replace parseFromString() with something like [fast-xml-parser](https://github.com/NaturalIntelligence/fast-xml-parser/tree/master/lib).  Double points if it has a callback where we can build XML objects without actually materializing DOM objects.  I haven't tried using an external library, yet.

Passes existing XML tests.

### DIFF
--- a/frameworks/projects/XML/src/main/royale/XML.as
+++ b/frameworks/projects/XML/src/main/royale/XML.as
@@ -34,20 +34,24 @@ package
 		 * The XML QName can be a significant percentage of an XML object size.
 		 * By retaining a lookup of QNames and reusing QName objects, we can save quite a bit of memory.
 		 */
-		static private var _nameMap:Object = {};
+		static private var _nameMap:Object = Object.create(null);
+		static private var _attrNameMap:Object = Object.create(null);
 		
 		static private function getQName(localName:String,prefix:*,uri:String,isAttribute:Boolean):QName{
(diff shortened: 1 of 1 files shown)


Harbs:
Looks good! Nice touch with `Object.create(null)`.

PR #1026, opened 2020-12-31
Speed up "x as y" and "x is y" where y is an interface

estanglerbm:
Speed up "x as y" and "x is y" where y is an interface, by flattening and caching results from checkInterfaces() inside Language.is().

I don't know how dynamic parent classes can be in Royale, but if they're not too dynamic, then here is my suggestion on how to speed up the incredibly-slow checkInterfaces() in Language.is().  It uses 2-dimensions of WeakMap, on browsers that support it, to flatten and cache the results of checkInterfaces().

To give you an idea about sizes, on a medium-complexity app using Spark and MX, interfaceMap grew to 234 entries (corresponding to leftOperand / classes), and from an informal survey, it appears that many (if not most or all) stayed below 22 entries per class (corresponding to rightOperand / queries-about-interface);  many had only 1 entry per class.

I know there is a desire to avoid "is" and "as" and use @royaleignorecoercion whenever possible, for performance.  That's fine.  But there are some places, like UIComponent, where this is causing a real problem.  (More on that in another PR.)

Existing code has some severe performance problems in some areas, due to checkInterfaces().  For example SimpleCSSValuesImpl.getValue() (in royale core) is called recursively up its parents, in parts of Spark / MX, and can take an incredible amount of time due to the "is IStyleableObject" required check.

The goal of this change is make "is" and "as" performant where it is really needed in order to be correct code.

### DIFF
--- a/frameworks/projects/Language/src/main/royale/org/apache/royale/utils/Language.as
+++ b/frameworks/projects/Language/src/main/royale/org/apache/royale/utils/Language.as
@@ -56,6 +56,12 @@ package org.apache.royale.utils
         static private var muler:Number;
         static private var zeroStr:String = String.fromCharCode(0);
         
+        /**
+         * Var for speeding up checkInterfaces()
+         */
+        static private var interfaceMap:Object = null;
+        static private var isInitInterfaceMap:Boolean = false;
+        
         //--------------------------------------
         //   Static Function
         //--------------------------------------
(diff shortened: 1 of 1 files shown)


Harbs:
This is a great optimization! This is something I wanted to do for a long time. Thanks!

Any reason you didn't also optimize class lookups?

FWIW, I'd also like to optimize Array checks in the compiler where `foo is Array` would get compiled directly into `Array.isArray(foo)`.

estanglerbm:
@Harbs Which class lookups?

I didn't see anything like that that showed up at the top of profiling.  But I may not be doing anything intense in that area, perhaps.

(Also, I could see checkInterfaces() itself using interfaceMap to speed up individual checks.  But it's already fast enough where it doesn't hardly register on profiling, at least for me.)

Harbs:
Never mind. My comment made no sense. I thought we were walking of the chain manually, but `instanceof` does that for us automatically.

PR #1028, opened 2021-01-03
Better error reporting for Vertical/HorizontalLayout

stanislaw89:


Harbs:
FWIW, `org.apache.royale.debugging.assert` is probably better for this kind of error reporting because it's stripped out in the production version.

carlosrovira:
Thanks Harbs! I wasn't aware of that, will take a look and see how to adapt Jewel to it.

Harbs:
Here's an example of its use:
https://github.com/apache/royale-asjs/blob/develop/frameworks/projects/Basic/src/main/royale/org/apache/royale/routing/RouteToState.as#L111

carlosrovira:
Hi Harbs,
I'm looking for a way to rewrite the `setGap` with assert and don't find a way that like me. Can you suggest one?
Thanks

Harbs:
How's this?
https://github.com/apache/royale-asjs/commit/dedc8d237a62bfe19598bb0370930f4b7971c960

Harbs:
Not sure if the check is needed for production as well. I was assuming not.

If yes, you should keep the `if` but drop the `else`.

carlosrovira:
yeah, we can assume that the developer reach production with a right value, so I think it's ok.
I was trying to figure if that initial if was needed. Thanks!

PR #1031, opened 2021-01-04
MX ADG: Fix scroll handler (under async scroll events) and optimize current implementation of ADG

estanglerbm:
Fixed scroll handler (under async scroll events) and optimize current implementation of MX ADG, especially with regard to header renderers.

The "deferred" logic for the scroll handler in ADG (in MX DataGridLayout) wasn't quite right, and under async scroll events in more recent browsers, it fired too often.  More room for improvement, but should be better.

**Even with the following ADG optimizations, scrolling is still not as smooth as it should be.**  In fact, with the way the scroll event is hooked, it really doesn't tolerate much rendering time.  Now, in release mode, it may be barely acceptable for grids with more than a few columns and rows.

Removed things and cached things in ADG header renderers to optimize.  They need more refactoring to be more Royale-like, but this is some progress (and is a lot faster).  Note that to work around very slow getStyle(), some header styles are not as dynamic.

Some simple optimization (fetch positioner into variable) in a few UIBase methods (x, y, setX, setY) to reduce the number of function calls (even though function calls are cheap in JS, these are high-impact methods).

Added override of validateDisplayList() to AdvancedDataGridHeaderRenderer so that AdvancedDataGridHeaderLayout had a way to call AdvancedDataGridHeaderRenderer.updateDisplayList() [which is protected].  Previously, it was being called through setWidth(), but that's the wrong place and even more expensive.

Moved the scrollTop workaround (for whatever Chrome bug) to after the loop, for DOM performance.

Did some naughty style caching (through the owner grid) in AdvancedDataGridSelectableItemRendererBead so that the first load of cached styles can be shared among the beads over all the renderers, for performance.  Hopefully that can be rewritten with faster stuff in a normal way later.  [Did it this way because I can't find a good place to push the styles from the owner grid to each bead as it gets added.]

### DIFF
--- a/frameworks/projects/Basic/src/main/royale/org/apache/royale/core/UIBase.as
+++ b/frameworks/projects/Basic/src/main/royale/org/apache/royale/core/UIBase.as
@@ -723,13 +723,14 @@ package org.apache.royale.core
         {
             if(!isNaN(_x))
                 return _x
-            var strpixels:String = positioner.style.left as String;
+            var pos:WrappedHTMLElement = positioner;
+            var strpixels:String = pos.style.left as String;
             var pixels:Number = parseFloat(strpixels);
             if (isNaN(pixels))
             {
-                pixels = positioner.offsetLeft;
-                if (positioner.parentNode != positioner.offsetParent)
-                    pixels -= (positioner.parentNode as HTMLElement).offsetLeft;
+                pixels = pos.offsetLeft;
+                if (pos.parentNode != pos.offsetParent)
+                    pixels -= (pos.parentNode as HTMLElement).offsetLeft;
             }
             return pixels;
         }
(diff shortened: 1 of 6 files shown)


Harbs:
Quickly looked at the changes and your comments. Some suggestions:

Reading styles in JS is very expensive mostly because it often triggers a forced reflow.

I assume you are already using the browser profiling tools to see where the bottleneck is, but it's worth mentioning.

Very often it's possible to use `requestAnimationFrame` to delay and batch process reading of styles to prevent extra forced reflows.

It's also probably possible to make getStyle smarter so it only falls back to reading browser values when necessary.

estanglerbm:
Several of the changes I made did involve DOM attributes that were causing forced reflow (get scrollTop is still an issue, but I don't see how to fully avoid it).  And yes, I saw them in the browser profiler.

requestAnimationFrame() was already in the scroll handler, and from my quick testing without it, seemed to be necessary for correct results.

But the getStyle() issue involves AS code that is recursively looking up the class hierarchy and parent chain (in AS classes).  It's not looking at the DOM.  (Though, I was going to suggest that's one way of making getStyle() faster;  it's not great, but may be better than the current getStyle().)  See the discussion on the dev list.

Harbs:
`getStyle` uses `ValuesManager.valuesImpl` which is usually `SimpleCSSValuesImpl` or some subclass of that. IIRC, `SimpleCSSValuesImpl` uses the styles object which is a reference to the `CSSStyleDeclaration` in the browser.

Yes. Let's discuss on the list.

PR #1033, opened 2021-01-04
Fix IBeadView for Spark DataGroup, and add missing "implements ILayoutView" to Spark GroupBase.

estanglerbm:
Reason that Spark GroupBase worked was because (JS) Basic ContainerView.get contentView() has a @royaleignorecoercion for ILayoutView, and GroupBase just happens to implement all of the methods of ILayoutView.

Also removed wrong IContentView from SWF CSS of Skinnable containers.


piotrzarzycki21:
@Harbs I think you should wait for Yishay with merge that. There were some reasons why this was left like that.

Harbs:
Sorry. It looked like an obvious change to me.

estanglerbm:
Just for reference, before I submitted this PR, I checked other parts, and I found that all of the equivalent classes (such as Basic GroupBase and Jewel GroupBase) were implementing ILayoutView.  The Spark GroupBase was the odd one.  Basically, ILayoutView is required for ContainerView.  The methods implementing ILayoutView were already in Spark GroupBase.  I've been running with the change for a month.

PR #1034, opened 2021-01-04
Under Spark, systemManager can be null in UITextFormat.measure()

estanglerbm:
For whatever reason, systemManager can be null in the context of UITextFormat.measure() while running under Spark.  (I think this was primarily in MX ADG, but I can't remember exactly.)

So this PR just makes a standalone measuringElement, in that case.

Perhaps there's a better way of dealing with it than this.  Or systemManager is supposed to not be null here?  Open to any ideas.

### DIFF
--- a/frameworks/projects/MXRoyale/src/main/royale/mx/core/UITextFormat.as
+++ b/frameworks/projects/MXRoyale/src/main/royale/mx/core/UITextFormat.as
@@ -574,11 +574,11 @@ public class UITextFormat extends TextFormat
         var measuringElement:HTMLSpanElement = measuringElementRef;
 
         if (!measuringElement) {
-            measuringElementRef = (systemManager as SystemManager).measuringElement;
+            measuringElementRef = (systemManager ? (systemManager as SystemManager).measuringElement : null);
             measuringElement = measuringElementRef;
             if (!measuringElement) {
                 var sm:SystemManager = systemManager as SystemManager;
-                if (sm.measuringElement == null)
+                if (!sm || sm.measuringElement == null)
                 {
                     measuringElement = document.createElement("span") as HTMLSpanElement;
                     //everything else is absolute position so should be above this element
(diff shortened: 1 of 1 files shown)


Harbs:
Probably worthwhile figuring out why it's null, but until then, we should probably merge this.
@aharui @greg-dove @yishayw Thoughts?

Harbs:
This Pr has been open a long time. I don't use MX classes. Should this be merged? Is it still needed?

frameworks/projects/MXRoyale/src/main/royale/mx/core/UITextFormat.as
```diff
@@ -587,8 +587,8 @@ public class UITextFormat extends TextFormat
                     //sm.measuringElement.style.display = "none"; // to try to keep it hidden
                     measuringElement.style.opacity = 0;
                     measuringElement.style["pointer-events"] = "none";
-                    sm.element.appendChild(measuringElement);
-                    sm.measuringElement = measuringElement;
+                    if (sm) sm.element.appendChild(measuringElement);
```
yishayw, line 590: So if sm is null then no measuring element is created?

yishayw:
If I understand correctly UITextFormat is used primarily by UITextField, which from a brief look is only used in charts in mx.charts.AxisRenderer. Generally speaking it looks like systemManager references are retrieved from the parent recursively until Application is reached, which has its systemManager initialized when the application is started. I haven't dived deep enough to guess when this would fail (didn't check popups for instance). To me it doesn't look like systemManager is ever supposed to be null, so I would prefer to at least open a bug on that before merging.

frameworks/projects/MXRoyale/src/main/royale/mx/core/UITextFormat.as
estanglerbm: In this function, "measuringElement" is what is being used, so that's being created.  The code above is just saying that sm.element and sm.measuringElement won't be updated, because, you know, !sm.

But yes, if !sm, then it is leaking "span" elements.  But I don't think !sm lasts forever (and I probably didn't want to update this.measuringElementRef because I wanted to allow sm to eventually be set, somehow).  I choose temporary leak over crash.  I didn't debug enough to determine why !sm.

estanglerbm:
UITextField is way more widespread in MXRoyale than just charts.

yishayw:
> UITextField is way more widespread in MXRoyale than just charts.

So I probably don't understand correctly... Where else is it constructed other than in AxisRenderer ?

estanglerbm:
UITextField is in ADG, and it's used by SWF UITextFormat, and other uses.

But actually, UITextFormat is the issue, not UITextField.

UITextFormat (which has the code in this PR) is used in UIComponent.measureText().  UIComponent.measureText() is used all over the place.  That's the greatest impact.

PR #1036, opened 2021-01-04
XMLNotificationTest: Added new callee parameter at end of notification callback function.

estanglerbm:
This PR goes with PR #969.

### DIFF
--- a/frameworks/projects/XML/src/test/royale/flexUnitTests/xml/XMLNotificationTest.as
+++ b/frameworks/projects/XML/src/test/royale/flexUnitTests/xml/XMLNotificationTest.as
@@ -71,7 +71,8 @@ package flexUnitTests.xml
         }
 
         private var tracking:Array;
-        private function trackChanges(currentTarget:Object, command:String, target:Object, value:Object, detail:Object):void{
+
+        public function trackChanges(currentTarget:Object, command:String, target:Object, value:Object, detail:Object, callee:Function = null):void{
             var trackingRecord:String=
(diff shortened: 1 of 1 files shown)


greg-dove:
Do you have any other changes locally that might not be added to this PR? When I tried switching on the XMLNotitficationTest, I could see all tests passing except the 'testReplaceNode' test. We could add [Ignore] to that test for now. I think you also mentioned at one point that there were issues with the representation of the expected results (iirc - I might be wrong) in the tests, so perhaps you have fixes for that? I'm happy to merge this as-is, but I wanted to check first if there were some additional local changes you had not included here... ? Thanks!

estanglerbm:
I don't have any more changes in XML (just some changes in other modules to support XML values).

As I mentioned, I didn't implement "nodeChanged" yet, and that's why the one test is failing (as well as some tests that I have).

I don't think it's going to be easy to implement "nodeChanged".

greg-dove:
Ok, cool, thanks for confirming. I did have that replaceNode test working in the original prototype swapping implementation I started for this (see: https://github.com/apache/royale-asjs/blob/a747d311b819f3263e2a24a27066ffb58a1ad8dd/frameworks/projects/XML/src/main/royale/org/apache/royale/language/NotificationXML.as#L143) , but not sure if anything I did there helps with this approach (and I had not considered bubbling at all when I started looking at that, it was simply swapping the runtime behaviour of instances if they had notification functions assigned). I'll merge this as-is, and switch on all the notification tests except the failing one - thanks again!

greg-dove:
> that's why the one test is failing (as well as some tests that I have).

btw feel free to add failing tests if you like. They can be marked with [Ignore] for now.

PR #1037, opened 2021-01-05
Initial version of Spark formatters and mx.globalization.

estanglerbm:
Initial version of a spark.formatters and mx.globalization.

mx.globalization is the replacement for flash.globalization.  Since Royale already has its own Formatters area, and this implementation of globalization relies on MX formatters, and Spark already relies on MX, MX seemed to be the best place for it.

This doesn't provide a true locale system at the moment, but it's a start.  There are also some limitations of the existing mx.formatters.


estanglerbm:
Can this be merged?

estanglerbm:
Updated to add Spark formatters to MXML.

yishayw (changes requested):
I think you can copy docs from flash docs, e.g. 

https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/globalization/DateTimeFormatter.html

estanglerbm:
Added (endless) docs to mx.globalization (flash.globalization).

estanglerbm:
Tested by running (and reviewing output of):

Issue #1022, opened 2020-12-30
Copy row as text option not shown when right click on AdvancedDataGrid row


pashminakazi:
In Royale, there's no option to copy text, select text, and copy row as text.
![image](https://user-images.githubusercontent.com/42200979/103359456-ca75ab80-4ad9-11eb-8fc6-4a459c0185a0.png)

In Flex:
When right-clicking on any cell there's an option to copy row as text
![image](https://user-images.githubusercontent.com/42200979/103359076-01978d00-4ad9-11eb-9980-41a20a936da0.png)
Paste copied row
**Attribute	Value
RICA	Y**

or
if right-click on any text copy options also shows
![image](https://user-images.githubusercontent.com/42200979/103359071-fcd2d900-4ad8-11eb-81c9-644eccb9aa4a.png)

estanglerbm:
Looks like custom right-click actions?

Code to format the data from the grid cells is fairly straight-forward.

I have an implementation of System.setClipboard() that I could clean up and check in, for the "Copy Row" action, if that would help.

The "Export to Excel" action would be via FileReference.save().

For the right-click action menu, I guess ContextMenu needs to be implemented for JS?

Might be easier to implement via buttons, right now.

pashminakazi:
@estanglerbm when will you check-in your commits for "Copy Row", It's urgent please help

estanglerbm:
@pashminakazi Oh, I didn't realize that you were waiting on me for something.

I checked in MX System.setClipboard() for the actual copy-text-to-OS-clipboard part, in PR #1038.

I'll post some example code for getting info out of an MX AdvancedDataGrid / Spark DataGrid.

pashminakazi:
@estanglerbm Thank u so much
I want to ask a question, 
Are you talking about ContextMenu Class or ContextMenu Event?

for right-click I have to add ContextMenu Event in any class which was in IterativeObject class in flex, if yes then in which class should I add this Event?

or you are talking about ContextMenu Class?

estanglerbm:
Here is some sample code that copies grid data to the clipboard in CSV format (copyValues), and some sample code that copies grid data to a file in CSV format (saveValues) for use in Excel:

[clipboard-example.as.txt](https://github.com/apache/royale-asjs/files/5769808/clipboard-example.as.txt)

Note that I've changed the code up from the original, and I have not compiled it, so YMMV.

[...]

PR #1038, opened 2021-01-05
Add MX System.setClipboard()

estanglerbm:
For issue #1022.

The way the JS code adds an element to the DOM and uses it for the copy may cause the content to show on the screen for an instant.  That part could be improved.


Harbs:
If you set the visible:false (or maybe display:none) that should solve the flashing.

PR #1046, opened 2021-01-11
Fix height too small for Spark TabBar tab labels on Chrome

estanglerbm:
Was fine on Firefox, but on Chrome, showed horizontal scroll bar, obscuring the tab labels.  Adding 1 to the height fixed the issue, but for this fix, added 2, as it seemed more appropriate (even number, and on Chrome, button seems to extend a few pixels lower).  The box-sizing style had already been applied appropriately, so there is some other reason why Chrome requires 1 more pixel.


Harbs:
@aharui was there a reason you originally picked the value of 18?

estanglerbm:
FYI, presentationModel.rowHeight is a fixed constant used in place of measuring actual row height in several places, such as VirtualListVerticalLayout and ListItemRendererInitializer.  So it shows up in Spark List and Spark TabBar and other places (not just Spark).  It makes a big difference for virtualized layouts, like for Spark List, that will take the available space and divide by this number.  (In ADG, things are modified a bit.)

Actually, I retract my change on this;  18 is fine for now (at least for the current font).  The problem in TabBar is that the button faces on the tabs add a border and such.  I'll fix it another way.

PR #1053, opened 2021-01-13
Completed 'todo' item: allow amf0 bead to upgrade to amf3 encoding

SemiConscious:
Adds functionality marked 'todo' in the source.

There are 2 data encodings used by AMF: AMF0 and AMF3. Royale supports AMF3 by default and AMF0 via the AMF0AMF3Context bead. There is a complication in the AMF0 encoding, which is that it can upgrade from AMF0 to AMF3 midway through a read (or write), and this change implements that for reads.

I also added an override of the base class reset function which switches the `switchedToAMF3` state variable back to false if `reset()` is called, so in the future if the code is extended to handle multiple reads it should call `reset()` at the beginning of each read.

This has been tested against the existing unit tests, and our (my employer's) production AMF service which requires every combination you can think of, including AMF0->AMF3 upgrade.

### DIFF
--- a/frameworks/projects/Network/src/main/royale/org/apache/royale/net/remoting/amf/AMF0AMF3Context.as
+++ b/frameworks/projects/Network/src/main/royale/org/apache/royale/net/remoting/amf/AMF0AMF3Context.as
@@ -37,8 +37,8 @@ package org.apache.royale.net.remoting.amf {
 			AMFBinaryData.installAlternateContext(AMF0AMF3Context);
 		}
 
-		//@todo :
 		private static const AMF0_AMF3:int = 0x11;
+		private var switchedToAMF3 = false;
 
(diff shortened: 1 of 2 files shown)


greg-dove:
Thanks for working on that. I only added he amf0 support a few months back and quite frankly was not expecting anyone else would need it! Much appreciated!

SemiConscious:
It's the least I can do! You people have done an amazing job on this - it
really is a complete life saver. Man years of coding would otherwise have
to have been thrown away.

I came a little unstuck trying to sort out a different AMF issue I came
across a couple of weeks ago. I submitted a pull request which got merged,
then had to be backed out because it broke the examples (didn't realise
there were any!) - and I'd appreciate your input.

I have created https://github.com/apache/royale-asjs/pull/1056 - I would be
very grateful if you would take a look.

All the best
Jim

PR #1054, opened 2021-01-13
Set AMF context to correct encoding

SemiConscious:
This fix does two things:

1. the current code will probably crash if a read of an AMF0 encoded message is attempted, as the reader assumes AMF3, and nothing I can see changes that, including using the AMF0AMF3Context bead. This fix ensures that when the encoding type (0 or 3) is read from the message, the subsequent read is carried out using that encoding.

2. the second benefit is customer experience. Without the bead, the existing code outputs an error that gives no clue to the nature of the underlying problem - this is the reason I have gone down this particular rabbit hole. With this fix, if an AMF0 message is read, this error is output when reading AMF0: `AMF0 support is unimplemented by default, supported via bead`. With the bead in use, this fix does nothing beyond simply allowing the data to be read correctly.


greg-dove:
Nice work - thanks again!

PR #1056, opened 2021-01-14
Use AMFNetConnection.call()'s command parameter for the AMF targetURI

SemiConscious:
@greg-dove @carlosrovira This is a second attempt to fix what I tried to fix in pull request https://github.com/apache/royale-asjs/pull/1023, which was subsequently reverted after problems with the examples. This fix addresses the targetURI issue without breaking existing code.

The problem this fix solves: in `AMFNetComponenent.onReadyStateChange()`, the message is encoded and sent without ever setting `messageBody.targetURI`, and the default value `"null"` is always used.

To refresh our memories, here is the explanation of the `NetConnection.call()` `command` parameter from the Adobe AS3 help:

> command:String — A method specified in the form `[objectPath/]method`. For example, the `someObject/doSomething` command tells the remote server to call the `clientObject.someObject.doSomething()` method, with all the optional ... arguments parameters. If the object path is missing, `clientObject.doSomething()` is invoked on the remote server.

My solution: allow the targetURI value to be passed in in the `AMFNetComponenent.call()` `command` parameter, in line with how it was invoked under flex in the past. This is stored in the `item` object and queued, and passed eventually to `AMFNetComponenent.onReadyStateChange()` where the value is used to correctly set `messageBody.targetURI`.

I cannot find any instances in the royale code base where a value of anything other than null is passed into this parameter, so existing functionality will not be affected by this change.

Unfortunately - I have been unable to get the `SampleAMFWebApp` test application to run for love or money, though it builds fine, so I have been unable to definitively test that the examples work. I have eyballed the code carefully. I put the certainty that this doesn't break anything at 90%, but I have run out of time to debug the web server given the current pressure I am under on porting our app. If you have a working setup with a functional `SampleAMFWebApp` it would be really great if you could confirm that all is well prior to merging.

FYI in my previous attempt I made a change to the `NetConnection.call()` design. This solution doesn't mess with that - I found I am able to address my issue by using `AMFNetConnection` instead of `NetConnection`.

### DIFF
--- a/frameworks/projects/Network/src/main/royale/org/apache/royale/net/remoting/amf/AMFNetConnection.as
+++ b/frameworks/projects/Network/src/main/royale/org/apache/royale/net/remoting/amf/AMFNetConnection.as
@@ -301,6 +301,7 @@ public class AMFNetConnection
             requestQueue.push(
                     {
                         url: url,
+                        targetURI: command,
                         responder: responder,
                         args: params
                     }
@@ -328,6 +329,7 @@ public class AMFNetConnection
             var actionMessage:ActionMessage = new ActionMessage();
             var messageBody:MessageBody = new MessageBody();
             sequence++;
+            messageBody.targetURI = call.item.targetURI;
             messageBody.responseURI = "/" + sequence.toString();
             messageBody.data = args;
             actionMessage.bodies = [ messageBody ];


greg-dove:
Thanks again, I will check this and probably merge it later today or tomorrow.

One thing that could be super helpful is if you could check replacing AMFNetConnection with the AMF0NetConnection code.
AMF0NetConnection was not actually intended to be separate, I just added something to get low level calls working with AMF0 when I needed that (it was actually low level actionscript, not Flex code that was being ported), but that code was ultimately intended to swap into AMFNetConnection. It includes the fix you added here and a few other things.
I had started trying to address various issues, but was not able to test broadly and had to set the work aside (for some time now, for similar reasons that you have expressed). Because I know others were using remoting, I was hesitant to make substantial changes to the original AMFNetConnection without being sure. But if you swapped AMF0NetConnection code into AMFNetConnection and everything worked, it could be a good verification of things, and because you 'have your head in this stuff' I would welcome your scrutiny of that.... only when/if you get the chance.
See comments in commit 5adce5d6f2d51d77489f8bd1442afa91438a877b

greg-dove:
If you get a chance to look at the AMF0NetConnection  (ignore the name, it could have been named FooNetConnection it was just what I used at the time because I was working on AMF0) code to replace what is currently in AMFNetConnection, that would be awesome. For now I will merge this - thanks again! I tested it against the SampleAmfWebApp. FYI I think the trick to getting that to run (for me at least) is to use Java 8 and not a later version (I know for  sure that it does not work with openJDK 11, but I was able to run it on my machine with: "C:\Program Files\Java\jdk1.8.0_181\bin\java.exe" -jar target/SampleAmfWebApp-0.9.8-SNAPSHOT-exec.war ).

SemiConscious:
Hey Greg - thanks for your help with my pull requests, and that's an
excellent idea regarding Java version - I will give it a shot.

I have just built my app with AMFNetConnection replaced by
AMF0NetConnection and it works out of the box. Would you suggest that I
stick with AMF0NetConnection? I was initially concerned by the way that
there were no events on AMFNetConnection unlike NetConnection ... then I
looked at the NetConnection code and realised they won't actually get fired
in JS-land - should that be a concern?

On a more general note - I have come to a bit of a crossroads in my port.
It currently has MX controls and UI. I seem to be hitting a lot of problems
with the MX controls in Royale, mainly unimplemented functionality. I see
that Carlos has previously recommended going with Jewel throughout, and I
can see many controls in the tour de jewel application that seem to be
working versions of things I don't have in MX, eg Tiles. I need to make a
decision whether to stay with MX and work around or fix issues as they come
up, or do a Jewel conversion. There are 225 mxml files (and another 200 as
files) in the project. Do you have a recommendation?

All the best
Rupert

carlosrovira:
Hi Rupert,

each path has its pros and cons. I'll try to explain both:
- Emulation path: You'll get to reuse most of your code base with little changes, so it's the fastest way to migrate. The problems are two: look and feel and themes are not done, so depending on your users that can be or not an issue. The other problem is that not all is emulated but will be more over time. Depending on your needs you can plan to help in that front, join others or contract people to help you.
- Jewel: You'll need to recode, so your codebase can't be reused the same as emulation, but in my experience for my migration I could change just little things in MXML or add some beads here and there and get my app to work. But that's my case, others can be very different. The pros: Jewel has look and feel (enough to be used in mainstream scenarios) and is more suited for the future (responsiveness, themes,...). The cons: still need some things to be done. For example, there's no Tree, Multiple selection, AdvancedDataGrid,...are some of the things still undone in Jewel.

HTH

PR #1060, opened 2021-01-22
Update TextInput.as

jmegonzalez:
Fixed issue when value is "true" the readOnly attribute is set to "true", instead of "false".


Laturine:
@greg-dove @Harbs @carlosrovira @aharui  Any particular reason this can't be merged in? It would be awesome to be able to type a date in DateField.

PR #1077, opened 2021-02-08
New bead BadgeWithMouseClick

mjesteve:
I have made the changes in separate commits in case the proposed examples do not adapt.

Hiedra.

### DIFF
--- a/examples/jewel/TourDeJewel/src/main/resources/jewel-example-styles.css
+++ b/examples/jewel/TourDeJewel/src/main/resources/jewel-example-styles.css
@@ -132,3 +132,59 @@ j|IconButtonBar
     IItemRenderer: ClassReference("itemRenderers.NavigationGroupExampleItemRenderer");
 }
 
+
+/* List Pills CSS */
+.pillListStyle{
+    IItemRenderer: ClassReference("itemRenderers.PillsListItemRenderer");
+    background: transparent !important;
+    border: none !important;
+}
+.pillListStyle>.jewel.item.hovered {
+    background: transparent;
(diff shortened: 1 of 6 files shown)


carlosrovira:
Hi Maria Jose, 
thanks for the contribution. About the bead commit, I see it completely right.
About the examples, I see examples I see various things:
- hardcoded styles. In this case the bead just add a click behavior. So don't understand the need of all the styles added to TDJ. Think that TDJ has already considerable weight, and we need to keep in diet as much as we can.
- Also I see examples in Button and Combo pages and additional renderer. Although is very valuable, I think the example should be just add the bead to one of the Badges already in place in the footer to show how it works.

I think is good to separate the commits to help the review, but for next PRs, can you separate "by functionality", I mean. One PR for the framework bead and other for the examples? In the current way, it helps review, but since the first is ok but the later not, I still can't merge it.

Thanks again for sharing the code! :)

mjesteve:
Hi Carlos, let's see if I understand you well ...
What I should have done is send a PR with the bead and another PR with the examples?

I create a new PR only with the bead and in another separate PR the small example that you tell me in ButtonPlayGround (I don't understand if I should simply add the bead to one of the existing buttons or I can create a new one, under the last one)

Sorry for the inconvenience.
Hiedra.

carlosrovira:
Hi,

yes one PR for the bead, and another for adding the new bead to one of the existing badges in TDJ.
That way we can merge valid units on the fly and since are different if one has something wrong, that will not affect the other(s).
Thanks!

PR #1079, opened 2021-02-13
jewel-daterangerestriction: Replace time calculation with appropriate filters per view state

brianraymes:
The current date range restriction bead is based on time instead of calendar days and doesn't work for each of the date chooser's other views (day, month, and year).

This change disregards time as that doesn't allow the ranges to be inclusive and adds filtering for each view.

The current restriction bead will not allow you to choose the same day for a start and end range. This resolves that issue.


carlosrovira:
Thanks for this fix Brian! :)

PR #1126, opened 2021-07-11
Add support for function instances to getQualifiedClassName().

estanglerbm:
Added support for function instances passed into getQualifiedClassName().  For example, getQualifiedClassName(getQualifiedClassName).

Did not add support for calling with "Function", i.e. literally calling getQualifiedClassName(Function).  In that case, it's supposed to return string "Function", but no idea how to differentiate between function instances and "Function" (both have type "function" internally, like so many other things) in JS side.

Also added some tests in Reflection for this, but commented out ones that won't work right now.

getQualifiedClassName(myfunc) returns a string with "::" in it, so I made special exception for that on the SWF side.  Currently, due to commit 6cce56f26414653c33cc79cc772b066400db3e5b, on the SWF side, it generally replaces "::" with ".".  I assume it's to not have to add code for making the JS side match.  Should this getQualifiedClassName() be different from Flex in that respect?

### DIFF
--- a/frameworks/projects/Reflection/src/main/royale/org/apache/royale/reflection/getQualifiedClassName.as
+++ b/frameworks/projects/Reflection/src/main/royale/org/apache/royale/reflection/getQualifiedClassName.as
@@ -38,8 +38,11 @@ COMPILE::JS{
 	{
         COMPILE::SWF
         {
+            var s:String = flash.utils.getQualifiedClassName(value);
+            if (s === "builtin.as$0::MethodClosure")
+                return s;  // don't replace ::
             //normalize for Vector:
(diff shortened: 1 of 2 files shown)


estanglerbm:
On the last point, for example:  Flex would return "mx.charts.chartClasses::ChartLabel", but Royale returns "mx.charts.chartClasses.ChartLabel".

estanglerbm:
Any thoughts?

greg-dove:
Are you ok if I look into this over the coming weekend? I'm going to struggle to look at it before then, I'm afraid.

estanglerbm:
No rush.

PR #1041, opened 2021-01-06
Reconcile Basic ContainerView for SWF and JS, and some SWF fixes

estanglerbm:
Reconcile Basic ContainerView for SWF and JS (emphasis on not changing JS side), and fix some issues on SWF side.

The SWF and JS sides were quite different in several aspects, which made things confusing (and probably incorrect on the SWF side).

The motivation was to make it easier to debug, and to fix some SWF runtime issues.

I carefully went through and made sure I reconciled properly, but anyone is welcome to double-check.  There should not be any changes in behavior on the JS side.  The changes look like a lot because code pieces had to be moved around and combined, and unneeded code was eliminated.

PR #1129, opened 2021-07-13
Several Container fixes for SWF target (not related to PR 1041)

estanglerbm:
Several Container fixes for SWF target (not related to PR 1041).  These are for issues that predate PR 1041.

That is, after the fix for PR 1041, this is a different set of fixes that are needed to run Containers under SWF.

NOTE:  This may expose (legitimate) problems in some Containers outside of Spark that are using Core LayoutBase but not marked as implementing ILayoutParent.  In that case, there will be an exception in LayoutBase.performLayout(), in either JS or SWF, for the host (Container).

### DIFF
--- a/frameworks/projects/Core/src/main/royale/org/apache/royale/core/LayoutBase.as
+++ b/frameworks/projects/Core/src/main/royale/org/apache/royale/core/LayoutBase.as
@@ -310,7 +310,6 @@ package org.apache.royale.core
 		 *  @playerversion Flash 10.2
 		 *  @playerversion AIR 2.6
 		 *  @productversion Royale 0.8
-		 * @royaleignorecoercion org.apache.royale.core.ILayoutParent
 		 * @royaleignorecoercion org.apache.royale.events.IEventDispatcher
 		 */
 		public function performLayout():void
--- a/frameworks/projects/Basic/src/main/royale/org/apache/royale/html/supportClasses/Viewport.as
+++ b/frameworks/projects/Basic/src/main/royale/org/apache/royale/html/supportClasses/Viewport.as
@@ -87,6 +87,9 @@ package org.apache.royale.html.supportClasses
 			{
 				contentArea = new c() as UIBase;
 			}
+
+			if (!contentArea)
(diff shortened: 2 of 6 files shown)


estanglerbm:
(Not technically just for Containers) Updated with change needed to use skinClass style in skin MXML with SWF target.

frameworks/projects/Core/src/main/royale/org/apache/royale/core/LayoutBase.as
```diff
@@ -310,7 +310,6 @@ package org.apache.royale.core
 		 *  @playerversion Flash 10.2
 		 *  @playerversion AIR 2.6
 		 *  @productversion Royale 0.8
-		 * @royaleignorecoercion org.apache.royale.core.ILayoutParent
```
Harbs, line 313: Why was it necessary to remove this?

estanglerbm: It wasn't strictly necessary.  I could put it back in, if that's better.  I took it out because there were hosts not implementing ILayoutParent (but were implementing getLayoutHost()), and this wasn't being caught anywhere on the JS side (only SWF side).  I thought it would be helpful if somewhere in LayoutBase it caught that, and this seemed the best place.  Since casting is a much cheaper operation now, I didn't think it would be bad for performance, but I haven't measured that.

yishayw:
Do these address the issues found by Josh?

https://github.com/apache/royale-asjs/pull/1041#issuecomment-877182238

estanglerbm:
@yishayw No, that's why I wrote "(not related to PR 1041)".

yishayw:
Looks good to me. @Harbs , I just noticed you were the official reviewer to this. I felt @estanglerbm 's reasoning for removing the ignore coercion statement was sound, but feel free to revert the merge if you want to take a deeper look if you haven't finished the review.

Harbs:
It's fine I guess. I spent a lot of time declaring whether coercions are needed or not.

It feels to me like we should have "debug coercions" which disappear in release mode...

PR #1135, opened 2021-07-21
Spark dynamic container layout

estanglerbm:
Improvements to Spark containers to better support more dynamic layouts and containers.

These significant changes had several goals:

- Properly handle layout when adding / removing elements.
- Fix various static layout issues.
- Simplify lots of legacy ad-hoc code for fixing layout problems, by finding what was really necessary.
- Fix some SWF target issues.

There's still lots to fix and improve, but these changes pass more tests.  (It might also break some, but hopefully not much.)

Some notable changes:

- Introduction of "displayView", which is either "contentView" or "skin".  While "contentView" is about holding elements, "displayView" is a more convenient way to work with what's displayed (which includes content and decorations of the skin, if applicable), especially for sizing.
- Containers delegate some sizing functions (i.e. measuredWidth) to the "displayView".  While beforeLayout() and afterLayout() do sync sizing of host and "displayView", it's not productive to keep trying to make sure they are 100% in sync in sizing all the time.  The "displayView" is the single source of truth, if possible.
- Introduction of interfaces ISparkLayoutHost (for "displayView") and ISparkContainer (for "layout").
- Changes for JS to SparkTextButtonItemRenderer to help measure before it's shown.

Future potential work:

- **SkinnableDataContainer**.  Still needs a fix or two to pass some tests that SkinnableContainer is passing.  In the meantime, for many uses, Spark DataGroup can be a more suitable choice.
- **MeasurementBead in MX / Spark**.  MX UIComponent should delegate measuredWidth / measuredHeight to UIBase.measurementBead (IMeasurementBead).  MX UIComponent currently gets IMeasuredBead functions, directly, through IFlexDisplayObject.
- **Application.setActualSize()**.  Due to Royale's lack of Flex's [commit-measure-layout cycle](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/mx/managers/LayoutManager.html) (and for good reason), Application.setActualSize() has to set explicit sizes on the contentView (displayView).  These are sticky, but unfortunately necessary, to avoid being clobbered by LayoutBase.layout() doing measurement.  This is just for Application without any explicit or percent sizes, which is a weird case.  Might find some better thing to do here, instead of making the explicit sizes.


yishayw (commented):
Hi @estanglerbm , before I try to dive into this can you comment on how this is different from ChilldResize branch?

yishayw:
Cool. As I said I’m really eager to have this merged because I know how problematic layout in emulation currently is. I just want to make sure we choose the most robust solution, and hopefully have it apply to both MX and Spark.

frameworks/projects/SparkRoyale/src/main/royale/spark/components/SkinnableContainer.as
```diff
@@ -991,9 +973,7 @@ public class SkinnableContainer extends SkinnableContainerBase implements IConta
         
         dispatchEvent(new Event("beadsAdded"));
         dispatchEvent(new Event("initComplete"));
-        if ((isHeightSizedToContent() || !isNaN(explicitHeight)) &&
-            (isWidthSizedToContent() || !isNaN(explicitWidth)))
-            dispatchEvent(new Event("layoutNeeded"));
```
yishayw, line 996: Why are these conditions no longer necessary?

estanglerbm:
Merged develop, except the parts not needed on spark-dynamic-layout.

PR #1147, opened 2021-08-31
Jewel: Cancel spinner auto increment/decrement when mousing away from buttons

brianraymes:
Jewel SpinnerMouseController has two bugs where auto increment/decrement timers become unstoppable if one click drags off of the buttons (click and hold, drag off button, then let go). This change cancels the timers, if running, when the mouse moves away from the buttons.


Harbs:
Looks good!

Is there a reason you used `mouseout` instead of `mouseleave`? I don't use Jewel, and I'm not sure what the structure is. Is there any chance of `mouseout` being fired when moving the mouse over a nested element?

brianraymes:
That's a great question. Primarily, because mouseleave was not modeled in MouseEvents, and I'm more familiar with Flex's older model. I'll take a look at mouseleave and test it out. If it works well, I'll add it in and update this PR. Thanks!

Harbs:
@brianraymes Are you still planning on making changes to this PR?

brianraymes:
> @brianraymes Are you still planning on making changes to this PR?

Yes! I will test out the other mouse event soon and update. Thank you for the reminder. ;-)

mjesteve:
Could we merge this change?

Harbs:
Done. :)

yishayw:
I think this is breaking the build

>     [java] <http://apacheroyaleci2.westus2.cloudapp.azure.com:8080/job/royale-asjs_jsonly/ws/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/controllers/SpinnerMouseController.as(161)>: col: 11 Error: Call to a possibly undefined method removeAllListeners through a reference with static type Timer.
     [java]
     [java]                              timer.removeAllListeners();
     [java]                                    ^

PR #1127, opened 2021-07-11
Various small fixes and changes

estanglerbm:
Various small fixes and changes.

PR #1150, opened 2021-09-19
Moved some (non-MX) changes from PR #1127.  Add support for XML data …

estanglerbm:
…in several areas, especially around sorting and grids.

Based on feedback from @harbs.


Harbs:
Thanks. I'm not going to be available to continue this discussion for the next week plus. I'll try to come up with some suggestions then.

Harbs:
Sorry I forgot to look at this.

Besides PAYG concerns, this does not look like it will work with minification without exports.

I'll try to see what I can do to make this work while addressing those two concerns...

Harbs:
Do you have a sample project for testing these changes that I can use for some refactoring?

PR #1153, opened 2021-09-28
Jewel VirtualComboBox only recognises itemRenderers assigned via className

mjesteve:
Those set through property "itemRenderer" in the mxml tag are not recognised.

It can be checked in the TDJ. If we change the mapping of the ItemRenderer in the VirtualComboBox of the VirtualListsPlayGround.mxml view from className to itemRenderer tag.

This is how it is now:

image
image

The result is:

image

If we change it to:
image

The result is:
image

I have located the problem in VirtualComboBoxPopUpView by overriding the function "get list". When creating the List associated to the dropdown/popup the itemRenderer assigned to the VirtualCombobox must be mapped to it.
(I have replicated the same function of ComboBoxPopUpView)

Are the modifications correct?

Thx.

### DIFF
--- a/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/views/VirtualComboBoxPopUpView.as
+++ b/frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/views/VirtualComboBoxPopUpView.as
@@ -23,6 +23,7 @@ package org.apache.royale.jewel.beads.views
     import org.apache.royale.jewel.List;
     import org.apache.royale.jewel.VirtualList;
     import org.apache.royale.jewel.beads.models.IJewelSelectionModel;
+    import org.apache.royale.jewel.supportClasses.combobox.VirtualComboBoxPopUp;
 
     /**
 	 *  The VirtualComboBoxPopUpView class is a view bead for the VirtualComboBoxPopUp.
@@ -58,6 +59,11 @@ package org.apache.royale.jewel.beads.views
                 _list = new VirtualList();
 				_list.addEventListener("beadsAdded", beadsAddedHandler);
             }
+
+			if((_strand as VirtualComboBoxPopUp).itemRenderer)
+			{
+				_list.itemRenderer = (_strand as VirtualComboBoxPopUp).itemRenderer;
+			}
             return _list;
         }
     }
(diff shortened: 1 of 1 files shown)


frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/views/VirtualComboBoxPopUpView.as
```diff
@@ -58,6 +59,11 @@ package org.apache.royale.jewel.beads.views
                 _list = new VirtualList();
 				_list.addEventListener("beadsAdded", beadsAddedHandler);
             }
+
+			if((_strand as VirtualComboBoxPopUp).itemRenderer)
```
piotrzarzycki21, line 63: I think this changes should be made in setter of strand if possible. I would also make it in following way:

```
var virtualComboBoxPopup:VirtualComboBoxPopUp = _strand as VirtualComboBoxPopUp;
if (virtualComboBoxPopup && virtualComboBoxPopup.itemRenderer)
{
     list.itemRenderer = virtualComboBoxPopup.itemRenderer;
}
```

piotrzarzycki21 (changes requested):

frameworks/projects/Jewel/src/main/royale/org/apache/royale/jewel/beads/views/VirtualComboBoxPopUpView.as
mjesteve: I looked at how it was fixed in the ComboBoxPopUpView and I thought I had "replicated" it but I didn't realise it was inside the "if(!_list)", I'll be more attentive next time (... the truth is that I was annoyed by the "list" but I didn't question it because I thought I had done it anyway :) )

I've included it in the "strand setter", as you indicated and verified the operation.

Hiedra.

mjesteve:
Hi,
After further debugging the code, I believe that the patched code should be moved to "get list()", as it is in the ComboBoxPopUpView.as
In the "set strand", it first executes the assignment in the parent class, "super.strand = value" which references "list" in several lines and each time it does the "get list()" it creates _list (new VirtualList()). I don't know why it doesn't save the value of _list from one call to another...

image

The code would look like this:

image

I have found that with this implementation the _list variable takes effective value on the first call.
What do you think?

Hiedra

mjesteve:
I have added the proposed modifications in the PR #1154

PR #1154, opened 2021-10-05
Jewel VirtualComboBoxPupUp custom item renderer 2

mjesteve:
Fixed previous PR #1153


piotrzarzycki21:
It looks like previous attempt is not the right way and it didn't work ?

piotrzarzycki21:
Let's merge this one - I believe if we are doing similar things in other places, let's proceed.

mjesteve:
Correct.
It worked with the itemRenderer recognition in the CSS (j|VirtualComboBox) but not with the direct mapping.

PR #1155, opened 2021-10-08
jewel ThreeCheckBox - first commit

mjesteve:
In this first proposal I have included in the _checkbox.sass files this new control (like the Switch component) but as we said, maybe it is better to create your own.

I have managed to get it to work with the full SDK compilation, including the option "-P option-with-sass-compile" but if I only compile "themes" and "Jewel" it doesn't export the css's correctly.

### DIFF
--- a/examples/jewel/TourDeJewel/src/main/royale/CheckBoxPlayGround.mxml
+++ b/examples/jewel/TourDeJewel/src/main/royale/CheckBoxPlayGround.mxml
@@ -93,6 +93,54 @@ limitations under the License.
 				</j:CardPrimaryContent>
 			</j:Card>
         </j:GridCell>
+
+		
+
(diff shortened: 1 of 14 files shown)


Harbs:
What do you think of changing the name to `TriStateCheckbox`. I think that's the standard name for this kind of component.

mjesteve:
Hi Harbs, sure, no problem, can I change it on my PR?

Are you looking at it?, how do you see it?, is there anything that could be done better?
Piotr was going to look at it (I don't know if he already did), we were commenting that maybe it would be good to separate the sass implementation since now I included it in the checkbox.sass (same as the swith). He was not clear about it, what do you think?

Harbs:
I'm not using Jewel and haven't tested it. After quickly reading the code, overall it looks good. The only thing that jumped out at me is I think `indeterminated` should probably be `indeterminate`.

https://www.dictionary.com/browse/indeterminate

mjesteve:
You are right, I will change it too.

mjesteve:
Is it okay for you to mix the branch?

PR #1156, opened 2021-10-14
Update MXRoyaleBase build.xml to match the one in MXRoyale.

estanglerbm:


greg-dove:
Thanks @estanglerbm I will check this and merge asap.

greg-dove:
Thanks @estanglerbm, very much appreciated. I actually started work on the MXRoyale split over a year ago, and had copied that ant file as it was at the time I started. I set the work aside and only came back to it recently, and didn't check for the updates to that file since then or do file compare to the current MXRoyale so missed it - thanks heaps for picking this up. Hopefully that will solve Maria's issue - again, thanks for that!

PR #1164, opened 2021-11-13
XML - Copy blank attribute values.

estanglerbm:
Undoing a change made by @harbs in commit 6664fb4bb5308b376ba269635a48e63ba729ba7b.

### DIFF
--- a/frameworks/projects/XML/src/main/royale/XML.as
+++ b/frameworks/projects/XML/src/main/royale/XML.as
@@ -1294,9 +1294,7 @@ package
 			xml.resetNodeKind();
 			xml.setNodeKind(getNodeKindInternal());
 			xml._name = _name;
-			if(_value){
-				xml.setValue(_value);
-			}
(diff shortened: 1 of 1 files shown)


Harbs:
This causes empty attributes to use more memory.

That commit made a very big difference in memory usage of XML. The values are not being set on purpose.

Harbs:
In fact, we're probably setting empty values in more places than we should.

Harbs:
Resolved in 4c4e9ee4e5f69d30b8bf25f3b236cf35aae70c4c

estanglerbm:
To be clear, you're talking about excessive memory usage because there are many empty strings (as part of parsing XML using DOMParser.parseFromString) besides attributes.

Harbs:
Exactly.

My use case made extensive use of `copy`, so fixing it in `copy` made a big difference. I should have done a better job on the optimization the first time around... 😉

PR #1195, opened 2022-05-07
Media project + audio player and video player for jewel

cristallium:
Hi,
Here is PR for adding "Media" project with 3 classes :
MediaElement,VideoElement and AudioElement

Then VideoPlayer and AudioPlayer for Jewel based on Media lib and updated TDJ with demo for Audio and Video player

I check compile all framework with Ant : OK
But not tested with Maven, I'm not sure it will be good


mjesteve:
what a job!!! As there is a lot of code, if you think it's ok, before mixing it, it would be good if one of the experts could review it. 
For now, what I can do is to check the maven compilation (I don't know much but I'll try).

cristallium:
Hi,

Yes of course, you can do what you think is most appropriate, I'm not an expert on the subject, I hope there will be no problem. If one of the experts could review it, it would be great, I didn't know that it was possible.

Many thanks for your proposal, effectively it would be great if you can check for Maven because I'm affraid that I miss some pom files

For the code, as there were JS types and prototype for most of web api, I essentialy wrap all the needed for playing audio and video by reading 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement. I added helper code on some function to be more friendly (like ScaleMode). I didn't use Audio web api because of limitation on small mp3 duration (it need to read all mp3 file with xmlhttprequest in a buffer, and decode buffer => this isn't good for large file).

If you could take me informed ...

mjesteve:
I have never worked with Audio and Video web either but I see it as a good start, it is useful.

mjesteve:
Media.swc or part of HTML5.swc?

mjesteve:
In TDJ, ScaleMode doesn't work for me, does it work for you?
In the code that I have downloaded the css classes are missing.

cristallium:
> Media.swc or part of HTML5.swc?

This is a very goot question. I'm not sure to have the right answer. If the _philosophy_ of Media.swc is to have the same methods as Flex then it's not the case, so HTML5.swc or HTMLMedia.swc would be better. Perhaps opinion from others would be necessary ? 

> In TDJ, ScaleMode doesn't work for me, does it work for you?

Yes it's working for me on FireFox and Chrome. Css is necessary for scaleMode. (I added it at the end of default.css of Jewel but sorry, I forgot to commit). 

Here are needed styles :

 ```
 .jewel.videoplayer.letterbox {
    object-fit: contain;
  }
  .jewel.videoplayer.zoom {
    object-fit: cover;
  }
  .jewel.videoplayer.stretch {
    object-fit: fill;
  }
```

mjesteve:
> Here are needed styles :
> 
> .jewel.videoplayer.letterbox {
>    object-fit: contain;
>  }
>  .jewel.videoplayer.zoom {
>    object-fit: cover;
>  }
>  .jewel.videoplayer.stretch {
>    object-fit: fill;
>  }

You have to put this in Jewel's sass, compile the Themes and upload the generated css. Have you ever done it?

cristallium:
> You have to put this in Jewel's sass, compile the Themes and upload the generated css. Have you ever done it?

I never do it. But I just tried, and after some trials and error, it's seems to be OK.
I commited defaults.sass and _videoplayer.sass

I had a quick look at HTML and HTML5 projects, I think you are right : Media.swc should be HTML5.swc as it's mainly HTML web api wrapping.
If needed do you want me to update PR ? (all classes move to HTML project and HTML5 project extends HTML) or do you move it when mixing code ?

mjesteve:
> > You have to put this in Jewel's sass, compile the Themes and upload the generated css. Have you ever done it?
> 
> I never do it. But I just tried, and after some trials and error, it's seems to be OK. I commited defaults.sass and _videoplayer.sass
> 
> I had a quick look at HTML and HTML5 projects, I think you are right : Media.swc should be HTML5.swc as it's mainly HTML web api wrapping. If needed do you want me to update PR ? (all classes move to HTML project and HTML5 project extends HTML) or do you move it when mixing code ?

Do not modify anything yet. My opinion is not conclusive. Let others give us their opinion.

cristallium:
Hi,
I think you can close this thread, I do another PR [#1200](https://github.com/apache/royale-asjs/pull/1200)

PR #1220, opened 2023-05-02
Fix case where sortable=false

cdouine39:
This change fixes the problem where sortable is still possible even though sortable=false is set in the mxml.

Please note that this does not fix the fact that all columns are sortable by default while they shouldn't. Haven't found a fix for that yet.

Let me know if the fix seems correct to you.


Harbs:
`org.apache.royale.utils.Language.string` should never be called directly.

Harbs:
`selectedColumn` also needs a type.

cdouine39:
@Harbs thanks for your feedbacks. I've done the updates. Let me know what you think.

Harbs:
`this.mx_controls_beads_AdvancedDataGridSortBead_adg.columns` does not look right. I'm guessing you got that from the JS? That should be just `adg.columns` (or `this.adg.columns`).

Harbs:
Thanks!

PR #1222, opened 2023-05-25
Improve tooltip display

cdouine39:
If the tooltip is very long, it was displayed on one line and sometimes we could not read the end. This change is to display it on several rows.

Old display : 
image

New display : 
image

### DIFF
--- a/frameworks/projects/Basic/src/main/royale/org/apache/royale/html/accessories/ToolTipBead.as
+++ b/frameworks/projects/Basic/src/main/royale/org/apache/royale/html/accessories/ToolTipBead.as
@@ -156,6 +156,10 @@ package org.apache.royale.html.accessories
 			tt.x = pt.x;
 			tt.y = pt.y;
 			host.popUpParent.addElement(tt, false); // don't trigger a layout
+			
+			// Display de tooltip on several rows if too long
+			tt.element.style.maxWidth = "300px";
+			tt.element.style.whiteSpace = "pre-wrap";
 		}
 
 		/**


piotrzarzycki21:
Hey @yishayw @greg-dove do you think this solution is reasonable ?

yishayw:
@cdouine39 , thanks for this pull request. One thing to remember when dealing with Basic beads (ToolTipBead is under the Basic library) is that we strive to keep basic functionality and minimal code. The goal is that applications written with Basic components will not contain unused code and will load as fast as possible. That's what's called the Pay As You Go (PAYG) principle, which is different from 'just in case' code which can have an adverse effect on performance.  Therefore, adding 2 lines of code, 'just in case' there may be a multi-lined tooltip does not comply with the PAYG principle. The solution could be to create a more specialized bead (which could still be part of the Basic library) to deal with multi-line tooltips (e.g. MultilineTooTipBead), which might extend TooltipBead to add these lines of code. That way applications that require multi-line tooltips can use that bead, but Basic applications with short tooltips will used the regular TooltipBad and will not load 'just in case' code they are not using. 

So to make a long story short, I suggest to add your code to a new bead which extends ToolTipBead, instead of changing ToolTipBead.

cdouine39:
@yishayw thank you for your answer. I like your idea of keeping things as simple as possible. 
I won't have time to do the change on Apache Royale though, so I'm gonna close this PR and will keep my change as a patch on our project since we do need this change.

yishayw:
Sure. IMO there is no need for you to create a patch. All you need to do is create CustomToolTipBead in your application source and use that instead of the one provided in Basic. That way you will not need to fork the repo with all the complications involved in that. Let us know in users/dev (or here) list if you have questions.
________________________________

PR #1223, opened 2023-05-25
Remove call to parent to create the tooltip

cdouine39:
When setting the tooltip, 2 different tooltip get created :  one when calling the parent, creating a mx.controls.beads.ToolTipBead and one in the method itself, creating a org.apache.royale.html.accessories.ToolTipBead


piotrzarzycki21 (commented):
I think there is missing event dispatch in toolTip property. When you call super there is being dispatched dispatchEvent(new Event("toolTipChanged")); - I believe you should add this here if you are not calling super.

cdouine39:
@piotrzarzycki21 thanks for your feedback. I've added the dispatchEvent, as you mentionned.

PR #1225, opened 2023-06-01
Change type of _tempIR

cdouine39:
When using datagrid with grouping, I had an error here since _tempIR is not a ILabelFieldItemRenderer but a AdvancedDataGridItemRenderer


piotrzarzycki21:
@greg-dove do you have time to approve that ? It looks reasonable.

greg-dove:
> @greg-dove do you have time to approve that ? It looks reasonable.

Sorry, @piotrzarzycki21 I can't do thorough review. I'm afk for at least 5 days while traveling, and so currently unable to compile or test anything. I also need to set up with new laptop etc when I arrive.

cdouine39:
I'm having second thoughts about this PR, I'm not sure it would work in all cases. I'll reopen if this really works after more testing.

PR #1232, opened 2023-09-21
fixes the embed tag under flex

jfmherokiller:
this fix can help with porting processes by allowing the libraries to work under apache flex.

### DIFF
--- a/frameworks/projects/MXRoyale/src/main/royale/mx/display/Bitmap.as
+++ b/frameworks/projects/MXRoyale/src/main/royale/mx/display/Bitmap.as
@@ -32,7 +32,7 @@ package mx.display
 
 		
 	COMPILE::SWF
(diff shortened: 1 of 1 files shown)


Harbs:
@yishayw @greg-dove Looks correct to me, but I have not been involved much in the mx porting effort.

Any reason not to merge?

greg-dove (changes requested):
This is really just an unimplemented stub in javascript to mimic the native flash class... so yes it probably makes sense to make the swf version at least compatible with native Bitmap base class by changing its ancestry. 
But in that case, imo you should also ensure the super call from the constructor is passing the constructor params - if you change that I will approve it - thanks:

`super(bitmapData, pixelSnapping,smoothing);`

greg-dove (approved):

PR #1268, opened 2026-04-12
Added theme setter

NechamiWormser:


Harbs:
Not sure how we're going to settle on theming, but OK for now.

PR #1282, opened 2026-04-16
Added Status

estib-prog:


estib-prog:
@Harbs Can you please review?

Harbs:
Why is the size in px instead of rem?

estib-prog:
Fixed.

frameworks/projects/Style/src/main/royale/org/apache/royale/style/skins/StatusSkin.as
```diff
@@ -0,0 +1,97 @@
...
+		// <span class="relative inline-block h-2 w-2 overflow-hidden animate-bounce rounded-full bg-[#3abff8] shadow-[0_2.8px_3.5px_-0.7px_rgba(15,23,42,0.22)] after:absolute after:left-[18%] after:top-[16%] after:h-[38%] after:w-[38%] after:rounded-full after:bg-white/60 after:blur-[1.4px] after:content-['']"></span>
+		// <span class="relative inline-flex h-2 w-2 overflow-hidden rounded-full bg-[#f87272] shadow-[0_2.8px_3.5px_-0.7px_rgba(15,23,42,0.22)] after:absolute after:left-[18%] after:top-[16%] after:h-[38%] after:w-[38%] after:rounded-full after:bg-white/60 after:blur-[1.4px] after:content-['']"></span>
+		override public function set strand(value:IStrand):void
+		{
+			super.strand = value;
+			if (_styles)
+				return;
+			var colorSet:ThemeColorSet = ThemeManager.instance.activeTheme.themeColorSet;
+			var baseColor:ColorSwatch = (host.theme && host.theme != "default") ? colorSet.getSwatch(host.theme, 600) : colorSet.getSwatch(ThemeColorSet.NEUTRAL, 300);
+			var heightWidth:String = computeSize(getSize(), host.unit);
```
Harbs, line 64: Using unit is probably wrong because this will not work for anything other than rem.

estib-prog: I used the `computeSize` function, so I assumed it would handle different units correctly. Is there a specific case where this breaks that I should account for?

PR #1284, opened 2026-04-19
added text input

chiyenna:


frameworks/projects/Style/src/main/royale/org/apache/royale/style/TextInput.as
```diff
@@ -0,0 +1,122 @@
...
+			{
+				element["maxlength"] = value;
+			}
+		}
+		public function get maxlength():Number
+		{
+			return _maxlength;
+		}
+		private var _min:String;
+		public function set min(value:String):void
```
Harbs: max and min are not supported in text inputs.

PR #1285, opened 2026-04-20
Range

estib-prog:


frameworks/projects/Style/src/main/royale/org/apache/royale/style/Range.as
```diff
@@ -0,0 +1,56 @@
...
+		public function Range()
+		{
+			super();
+		}
+		
+		COMPILE::JS
+		override protected function createElement():WrappedHTMLElement
+		{
+			var elem:WrappedHTMLElement = super.createElement();
+			type = "range";
```
Harbs, line 28: This should go in the constructor instead.

estib-prog: I fixed it.

frameworks/projects/Style/src/main/royale/org/apache/royale/style/skins/RangeSkin.as
```diff
@@ -0,0 +1,241 @@
...
+			_inputStyles = [
+					new Position("absolute"),
+					new Inset(0),
+					new Cursor("pointer"),
+					new OpacityStyle(0),
+					new HeightStyle("100%"),
+					new WidthStyle("100%")
+				];
+		}
+		private function getRangeSize():Object
```
Harbs: Don't use untyped objects. Each component should be calculated explicitly.

frameworks/projects/Style/src/main/royale/org/apache/royale/style/skins/RangeSkin.as
```diff
@@ -0,0 +1,241 @@
...
+			return _strand as Range;
+		}
+
+		override public function set strand(value:IStrand):void
+		{
+			super.strand = value;
+			if (_styles)
+				return;
+			_styles = [
+					new WidthStyle("400px"),
```
Harbs: Why are sizes in px?

estib-prog:
Fixed.

PR #1287, opened 2026-04-23
Added Loading spinner

GittyBlau:
@Harbs Can you review this please?


frameworks/projects/Style/src/main/royale/org/apache/royale/style/Loading.as
```diff
@@ -0,0 +1,41 @@
...
+	public class Loading extends StyleUIBase
+	{
+		public function Loading()
+		{
+			super();
+		}
+
+		private var _animation:Object;
+
+		public function get animation():Object
```
Harbs: This should be a string

GittyBlau: Done.

